PR #1, opened 2018-05-14
Add Travis config

gschueler:


clofresh:
:+1:

gschueler:
I added a deploy script, travis should run it on tags, and branches looking like "3.0.x".

the mechanism is:

on a git tags like `#.#.#` or `#.#.#-GA`:
- docs are built as GA and also deployed as latest version in `docs/`

on a git tags like `#.#.#-maint` or `#.#.#-maint-2018-05-12`:

- it is deployed as a top level version, but not used as the docs/ (latest) version, this will allow updating existing versions, or releasing an older rundeck version without making it the "latest" docs version

on a branch
- branches formatted like "3.0.x" are deployed as "3.0.x-SNAPSHOT", so updates to that branch will be deployed as they are added

PR #5, opened 2018-06-02
Add 3.0.0 alpha1 notes

gschueler:
@clofresh i added a static link in the index.md to the alpha notes, but if you want to do that another way that's fine, e.g. put it as the 3.0.x-SNAPSHOT version notes.

Alternately, we maybe we just link to the RELEASE.md file on github under the v3.0.0-alpha1 tag for Rundeck ? this url: https://github.com/rundeck/rundeck/blob/v3.0.0-alpha1-2/RELEASE.md


clofresh:
@gschueler I think having it in the docs repo is a good approach, so you could tweaks to it after the release. Maybe to avoid duplication, the code's README.md could link to this page instead of having a separate RELEASE.md?

PR #11, opened 2018-06-08
Docs for rundeck pro encryptable properties.

sjrd218:


en/administration/configuration/encryptable-properties.md
```diff
@@ -0,0 +1,24 @@
...
+
+After downloading the zip distribution of the latest verison, you will unzip the archive to your desired location. The unzipped Jasypt folder has a bin folder inside it. Navigate to that location, and use the encrypt.sh or encrypt.bat script to run the following command: `./encrypt.sh input="binder123" password="1PwdToBindThem$"`
+
+The script will print various information. You are interested in the OUTPUT section which will have a value that looks like: `axxXQWvbRT3pBNqhpsM4wxNBzJc6gSPw` (note that it will not be this value)
+
+You will take the encrypted string: `axxXQWvbRT3pBNqhpsM4wxNBzJc6gSPw` and use it in your rundeck-config.properties file like this:
+`rundeck.security.ldap.bindPassword=ENC(axxXQWvbRT3pBNqhpsM4wxNBzJc6gSPw)`
+
+### Decrypting rundeck-config.properties
+To decrypt the encrypted properties in your rundeck-config.properties file you will need to set the environment variable `RUNDECK_PROP_DECRYPTER_PWD` with the value of your master password before starting Rundeck.
```
clofresh: Wouldn't this be difficult to set without having the environment variable in the clear? You'd have to put put it in the initd script that everyone can already read

clofresh: I guess you could have it entered interactively by the person starting the app. Something like

```
# Check that it's an interactive session
if [[ -t "0" || -p /dev/stdin ]]
then
    read -sp "Enter master password: " RUNDECK_PROP_DECRYPTER_PWD
    export RUNDECK_PROP_DECRYPTER_PWD
else
  echo "Must run interactively to be able to enter master password"
  exit 1
fi
```

sjrd218: I was using this location on this jasypt site as the source for the technique: http://marker.to/vEzIPO 
My thought was that the script to start rundeck would probably be run by a tool like chef or puppet which have ways to store and decrypt sensitive information like this. I think you bring up a good point. Maybe we should ask potential users of this feature how they want to handle the master password.

clofresh: yeah, good call :+1:

PR #20, opened 2018-06-21
Refactor user jobs

clofresh:
splits up the job user manual page into runnings jobs, creating jobs, creating job workflows, defining job options and job plugins


clofresh:
going to merge so that the html is reviewable, but people can still leave feedback on this PR

gschueler:
notifications section has the "Progress Badge (Pro)" but i think that should be in Workflow step, and Log filter, not notification:

http://rundeck.org/3.0.x-SNAPSHOT/manual/job-plugins.html#notifications

gschueler:
Orchestrator plugins might go under Nodes section.  Workflow Strategy plugins probably could go in the Jobs Plugins.  Log Filter plugins as well?

PR #48, opened 2018-08-01
Fix #41 update auth instructions

### LINKED ISSUE #41: authenticating-users is wrong for 3.0
includes mention of `web.xml` which is no longer usable

http://rundeck.org/docs/administration/security/authenticating-users.html

gschueler:


ProTip:
For the RPM section should we mention that it applies to DEB as well?

PR #51, opened 2018-08-02
Update Oracle configuration guide.

ProTip:


gschueler:
I will schedule this for 3.0.1, since it will not apply to 3.0.0

PR #123, opened 2018-10-01
Replaced forwardticks with apostrophes

martinbydefault:
Right now it's not possible to copy&paste and execute the command because forward ticks aren't syntactically correct and the execution raises an error.


gschueler:
thank you!

PR #125, opened 2018-10-02
Update authentication.md for rpm/deb

thehackadmin:
Examples of terms should be consistent.  I used "myloginmodule" consistently to remove confusion.  
Also, added 2 more configurations that was necessary in my case... I assume that others using a deb/rpm install will also need this information.

### DIFF
--- a/en/administration/security/authentication.md
+++ b/en/administration/security/authentication.md
@@ -162,7 +162,7 @@ You can simply specify the system properties on the java commandline:
 ~~~~~~ {.bash}
 java -Drundeck.jaaslogin=true \
      -Dloginmodule.conf.name=jaas-ldap.conf \
-     -Dloginmodule.name=ldap \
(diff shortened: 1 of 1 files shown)


gschueler:
this info seems out of date.  we no longer use web.xml for rundeck 3.  Also, the profile should not be edited, instead the `/etc/sysconfig/rundeckd` or `/etc/defaults/rundeckd` should be edited.

thehackadmin:
Those are interesting comments... I did have 2 changes here, one that changes the examples to use the same text.  That would still be helpful.  

As for the web.xml file... perhaps my config is less than ideal.  How is the AD user account name set if not there?

As for the profile vs rundeckd files... I will have to look into that.  Again, maybe my config is not ideal.

gschueler:
Closing in favor of https://github.com/rundeck/docs/pull/208

PR #185, opened 2018-12-22
Update relative paths and html file names to link directly via absolute file path

skywalke34:
Also fixed ~90 broken URL links to resolve correctly within documentation.


gschueler:
@skywalke34 this PR needs to be redone..you need to make a pr to merge your branch "Readme.md" into 3.0.x

PR #207, opened 2019-01-25
Cluster config page enhancement

emiliohh:
With these changes, it should be easier to configure a Rundeck Cluster


plambert:
The product was renamed a few months ago from Rundeck Pro Cluster to Rundeck Enterprise; should we use that name in the documentation?

PR #255, opened 2019-05-21
Upgrading guide for Rundeck 3.1

gschueler:
WIP:

Please comment with Rundeck 3.1 upgrade guide information: Breakages, configuration changes, install requirements, etc.

### DIFF
--- a/en/upgrading/toc.conf
+++ b/en/upgrading/toc.conf
@@ -1,2 +1,3 @@
+1:upgrading-to-rundeck-3.1.md:Upgrade to Rundeck 3.1
 1:upgrading-to-rundeck3.md:Upgrade to Rundeck 3.X
 1:upgrading.md:Upgrade Guide (2.x)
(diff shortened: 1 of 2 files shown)


sjrd218:
New configuration: `rundeck.security.requiredRole=required_role`

A user must have the role specified in `rundeck.security.requiredRole` to be able to login if this configuration value is set.

ProTip:
Changes to the Docker image to make it compatible with OpenShift 4+ may break downstream image builds.

gschueler:
@sjrd218 is this something that would break an in-place upgrade? I thought it only was in effect if they add that configuration?

gschueler:
@ProTip can you write those instructions here? or is the description of that PR sufficient?

sjrd218:
@gschueler You are correct. It only checks the value if the configuration is set. Since it won't be set by default it shouldn't break any in-place upgrades. It would only affect people who intentionally add that configuration.

gschueler:
@ProTip i added the docker openshift content, lmk if it needs to change

ProTip:
@gschueler Looks good to me.

PR #275, opened 2019-07-01
Fix Autocomplete Bug

mishingo:
- Previous issue of non functional autocomplete doc search has been fixed


en/docs-autocomplete.js
```diff
@@ -71,11 +92,11 @@ function setupAutocomplete(site){
 jQuery(function(){
 	
 
-	jQuery.getJSON(window._pageRelpath+'site.json',function(data){
+	jQuery.getJSON('/site.json',function(data){
```
gschueler: this isn't going to work, i think the relative path is necessary. /site.json will resolve to the site base url

en/templates/html.template
```diff
@@ -12,42 +12,21 @@ $if(date)$
...
+<script defer src="https://use.fontawesome.com/releases/v5.3.1/js/all.js"></script>
+<script
+src="https://code.jquery.com/jquery-2.2.4.min.js"
+integrity="sha256-BbhdlvQf/xTY9gja0Dq3HiwQF8LaCRTXxZKRutelT44="
+crossorigin="anonymous"></script>
+<script
+src="https://code.jquery.com/ui/1.11.3/jquery-ui.min.js"
+integrity="sha256-xI/qyl9vpwWFOXz7+x/9WkG5j/SVnSw21viy8fWwbeE="
+crossorigin="anonymous"></script>
+<script src="/jquery.autocomplete.min.js"></script>
```
gschueler: the relpath needs to be inserted into the path here
```suggestion
<script src="${relPath}jquery.autocomplete.min.js"></script>
```

en/templates/html.template
```diff
@@ -12,42 +12,21 @@ $if(date)$
...
+<script src="/docs-autocomplete.js"></script>
```
gschueler:
```suggestion
<script src="${relPath}docs-autocomplete.js"></script>
```

PR #279, opened 2019-07-19
SQLRunner script inline enhancement 

jtobard:
Documentation for enhancement to sqlrunner to accept inline script and variables.
rundeckpro/rundeckpro#646


gschueler:
@jtobard does it support option expansion in the variables list?

jtobard:
Example of options usage on variables added.

PR #287, opened 2019-08-01
Fix Scroll on sidebar and column content

mishingo:


gschueler:
let's verify this on the snapshot published docs, and if it works we can redeploy the 3.1.0 docs

PR #329, opened 2019-10-16
WIP: add mermaid for graphs

gschueler:
Attempt to use [mermaid](https://knsv.github.io/mermaid/#/) to replace Markdeep that was used before for ascii diagrams.

Unfortunately, the rendering seems to cut off some of the text as seen here: 

Screen Shot 2019-10-16 at 6 32 20 AM

For reference, here is the markdeep svg rendering as seen in the [3.0 docs](https://docs.rundeck.com/3.0.26/manual/content-converters/index.html)

Screen Shot 2019-10-16 at 6 34 37 AM

### DIFF
--- a/docs/.vuepress/components/mermaid.vue
+++ b/docs/.vuepress/components/mermaid.vue
@@ -1,0 +1,19 @@
+
+<template>
+  <div class="mermaid">
+    <slot></slot>
+  </div>
+</template>
+
+<script>
+export default {
+  beforeMount() {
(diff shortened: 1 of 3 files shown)


ProTip:
@gschueler That's awesome. Shame the mermaid markdown-it plugin hasn't been updated in 2 years :/

fdevans:
Is this still something we want to include or should we cancel this PR?

gschueler:
i'll close it, if necessary we can revisit it

PR #363, opened 2019-11-26
Node Healthchecks

gschueler:
Documentation for Node Healthchecks

- [ ] add docs for specific health check plugins (command & script)


jessemarple (commented):
In setup.js, should we update the RUNDECK_VERSION_FULL from 3.1.. to 3.2?

gschueler:
@jessemarple i guess so, in the 3.2.x branch.  I asked @ProTip for more guidance, i wasn't sure what the behavior was

PR #406, opened 2020-01-06
Improvement to include execution context variables

fgutierrezz:
will be good to also include execution context variables as shown here "Execution data"

https://docs.rundeck.com/docs/developer/05-notification-plugins.html#execution-data


g3nsvrv (approved):
Looks Good

PR #409, opened 2020-01-08
Clean install when previous version is present.

fgutierrezz:
Will be good to include step to complete remove Rundeck ( clean install )


jessemarple (approved):
LGTM

PR #415, opened 2020-01-23
Include a note about java version new OS like debian 10 auto install Ja…

fgutierrezz:
…va 11

If java is not installed in for example Debian 10 and rundeck repository is set. It will automatically install OpenJDK11. 

Example: 

# apt-get install rundeckpro-cluster
Reading package lists... Done
Building dependency tree       
Reading state information... Done

The following NEW packages will be installed:
java-common openjdk-11-jre-headless >>>>>>>

0 upgraded, 5 newly installed, 0 to remove and 2 not upgraded.
Need to get 366 MB of archives.
After this operation, 514 MB of additional disk space will be used.
Do you want to continue? [Y/n]

### DIFF
--- a/docs/administration/install/linux-deb.md
+++ b/docs/administration/install/linux-deb.md
@@ -20,6 +20,25 @@ sudo apt-get update
 sudo apt-get install rundeck
 ```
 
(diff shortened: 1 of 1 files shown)


jessemarple (approved):
lgtm

PR #417, opened 2020-01-24
Update upgrading.md

jplassnibatt:
should not be removed all from server/lib/ cause the oracle driver might be there.


fgutierrezz:
May help removing only the rundeck*core files from server/lib folder

Example

```
# rm -rf  server/lib/rundeck-core* server/sbin/ tools/ var/.first* var/resource* var/.install_complete-missing-ver  var/tmp/ libext/ server/exp/;
```

fgutierrezz (approved):

PR #432, opened 2020-02-26
About using Multiple tabs in the browser

fgutierrezz:
Some customers have reported having unexpected issues when using multiples tabs.


jplassnibatt (changes requested):
@fgutierrezz Could you move it to the GUI section, please?

jplassnibatt (approved):
ok

PR #436, opened 2020-03-02
Improved Static authentication tokens

danielabbatt:
I spent an hour or so failing to work with the static token properly because I didn't realise you could also specify the roles applied to the token and be able to use the format
username: token_string,role1,role2
I was using this with docker and automated testing so wanted the tokens in there with roles similar as if the admin had created the token in the UI


gschueler (approved):

gschueler:
Thank you for the improvement!

PR #445, opened 2020-03-06
Job Restart Failed nodes plugin

ltamaster:


gschueler (approved):
Approved, but I think we should consider renaming the plugin to "Retry Failed Nodes" plugin (instead of Restart), since it requires the Retry option to be enabled

PR #450, opened 2020-03-12
Fixes broken link in API documentation - ACL Policy

jessemarple:


fdevans:
Doesn't seem like the link change is correct unless it will get translated on the backend to:
/docs/manual/document-format-reference/aclpolicy-v10.html

PR #455, opened 2020-03-23
Update version-3.2.4.md

fdevans:
Updated Overview for 3.2.4 release notes.

### DIFF
--- a/docs/history/3_2_x/version-3.2.4.md
+++ b/docs/history/3_2_x/version-3.2.4.md
@@ -1,6 +1,12 @@
 # Release 3.2.4
 
 Name: <span style="color: chartreuse"><span class="glyphicon glyphicon-paperclip"></span> "nacho chartreuse paperclip"</span>
+## Overview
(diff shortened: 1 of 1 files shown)


docs/history/3_2_x/version-3.2.4.md
```diff
@@ -1,6 +1,12 @@
 # Release 3.2.4
 
 Name: <span style="color: chartreuse"><span class="glyphicon glyphicon-paperclip"></span> "nacho chartreuse paperclip"</span>
+## Overview
+This release brings some impressive new performance enhancements to the Jobs Listing page, more granular ACL options and the ability to Retry a job on failed nodes. Version 3.2.4 will offer an 80% improvement in performance, especially for those projects with a large number of jobs.
```
ProTip, line 5: Is the 80% performance increase is specific to job listing? The last bit makes it sound like this may be a general performance boost of 80%.

jessemarple (approved):

docs/history/3_2_x/version-3.2.4.md
fdevans: It's just for the job listing.  You're right we could be more explicit with the wording.

PR #480, opened 2020-04-15
Resolution edits for Issue #326

fdevans:
Closes Issue #326  by adding the 4 sections referenced.

### DIFF
--- a/docs/manual/document-format-reference/job-yaml-v12.md
+++ b/docs/manual/document-format-reference/job-yaml-v12.md
@@ -200,7 +200,11 @@ value of `failed`. Allowed values:
 
 [`notification`](#notification)
 
-: Job result notifications
+: Job result notifications.
+
+[`orchestration`](#orchestration)
+
+: Orchestration Plugin configuration to determine node processing order.
 
 [`plugins`](#plugins)
 
@@ -894,6 +898,28 @@ Example:
 ```
 
 - For more information about the Webhook mechanism used, see the chapter [Integration - Webhooks](/manual/04-jobs.md#webhooks).
+
+### Orchestration
+
+Defines a Orchestrator Plugin that can be used to determine the order in which nodes are processed.
+
+`type`
+
+: The type identifier of the Orchestrator plugin
+
+`configuration`
+
+: Contains configuration attributes for the plugin.  Values will vary depending on Orchestrator plugin.
+
+Example:
+
+```yaml
+    orchestrator:
+    configuration:
+      attribute: sort-attr-on-node
+      sort: highest
+    type: orchestrator-highest-lowest-attribute
+```
 
 #### plugin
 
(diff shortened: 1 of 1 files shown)


docs/manual/document-format-reference/job-yaml-v12.md
```diff
@@ -833,6 +845,9 @@ Defines a notification for the job. You can include any of `onsuccess`, `onfailu
 
 `onsuccess`/`onfailure`/`onstart`/`onavgduration`/`onretryablefailure`
 
+::: Note: `onavgduration` also requires the following attribute set at the same level as `notification`
```
ProTip:
```suggestion
::: tip
`onavgduration` also requires the following attribute set at the same level as `notification`
:::
```

ProTip: This is the tip box format:
image

docs/manual/document-format-reference/job-yaml-v12.md
```diff
@@ -833,6 +845,9 @@ Defines a notification for the job. You can include any of `onsuccess`, `onfailu
...
+`notifyAvgDurationThreshold` - Add or set a threshold value to the avg duration in order to trigger this notification. Options: - percentage => eg: 20% - time delta => eg: +20s, +20 - absolute time => 30s, 5m Time in seconds if you don't specify time units Can include option value references like ${option.avgDurationThreshold}.
```
ProTip, line 851: It looks like this should be above the note? I believe the following section describes the keys under `onsuccess`/`onfailure`/`onstart`/`onavgduration`/`onretryablefailure`, but `notifyAvgDurationThreshold` is its own top-level key.

fdevans: yes, but it's only added in one case of the 5 choices above.  I waffled between both options, but the note text is meant to guide the user to add `notifyAvgDurationThreshold`

docs/manual/document-format-reference/job-yaml-v12.md
```diff
@@ -186,13 +186,25 @@ value of `failed`. Allowed values:
...
+`nodesSelectedByDefault`
+
+: Boolean option that determines if Nodes (based on node filter) are selected automatically or if the user running the job must select nodes from the results. (true | false)
+
 [`notification`](#notification)
 
-: Job result notifications
+: Job result notifications.
+
+[`orchestration`](#orchestration)
```
gschueler: since the section is actually called `orchestrator` i think this and the header section later should be renamed to "Orchestrator"

docs/manual/document-format-reference/job-yaml-v12.md
fdevans: Do we need changes here or can we see how it works as written?

gschueler (approved):

docs/manual/document-format-reference/job-yaml-v12.md
ProTip: LGTM. The tip could be changed to `warning` if we want to raise the visibility of it.

ProTip (approved):

PR #493, opened 2020-04-21
Add alias plugin, aliases for some docs

gschueler:
caveats: this plugin doesn't seem to work in "dev" mode, so only when the production build is created do aliased paths get created.  We can try it and see how it performs when deployed

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -58,6 +58,7 @@ module.exports = {
     })
   },
   plugins: [
+    ['alias'],
     [
       'autometa', {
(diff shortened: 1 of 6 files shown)


gschueler:
FYI: it seems that the redirects it puts in place aren't working correctly.  

e.g. I added an alias of `api/index.html` for the `api/rundeck-api.md`. 

But if you visit https://docs.rundeck.com/3.2.x/api/index.html it redirects to `https://docs.rundeck.com/api/rundeck-api.html` which is incorrect. It is not using the base url.

PR #514, opened 2020-05-28
Update project-motd.md

jplassnibatt:
rd command fixed, add gui pics from https://github.com/rundeck/docs/pull/513

### DIFF
--- a/docs/administration/projects/project-motd.md
+++ b/docs/administration/projects/project-motd.md
@@ -5,21 +5,24 @@ The motd content is interpreted as [markdown](http://commonmark.org/help/) text 
 
 ## Graphical Interface
 
-![Figure: Project motd](~@assets/img/fixme.png)
+![Figure: Project motd](~@assets/img/motdgui01.png)
+
+![Figure: Project motd](~@assets/img/motdgui02.png)
+
+![Figure: Project motd](~@assets/img/motdgui03.png)
 
 ## CLI Usage
 
 Update the motd with some text.
 
 ```bash
-rd projects motd put -p MyProject -t "This is the **motd** for MyProject"
+rd projects readme put -p MyProject --motd -t "This is the **motd** for MyProject"
 ```
 
 You might want to set the project home page to show the motd.
 
 ```bash
-rd projects configure set -p MyProject -- \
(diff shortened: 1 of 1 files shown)


docs/administration/projects/project-motd.md
```diff
@@ -5,21 +5,24 @@ The motd content is interpreted as [markdown](http://commonmark.org/help/) text
```
fdevans, line 5: If we are going to make changes let's improve this introduction.  "MOTD" should be described as "Message Of The Day" before using the acronym everywhere else.  Improve context of why and how it's used, etc.

docs/administration/projects/project-motd.md
```diff
@@ -5,21 +5,24 @@ The motd content is interpreted as [markdown](http://commonmark.org/help/) text
...
 ## CLI Usage
 
 Update the motd with some text.
 
 ```bash
-rd projects motd put -p MyProject -t "This is the **motd** for MyProject"
+rd projects readme put -p MyProject --motd -t "This is the **motd** for MyProject"
 ```
 
 You might want to set the project home page to show the motd.
```
fdevans: Let's take out the "You might want to..."  That's poor documentation practice and change it to clearly explain what the next set of code is doing.

docs/administration/projects/project-motd.md
```diff
@@ -29,13 +32,13 @@ rd projects configure set -p MyProject -- \
     PUT /api/13/project/myproject/motd.md
     Content-Type: text/plain
 
-    This is the **motd** for MyProject
+    "This is the **motd** for MyProject"
 
 ## Filesystem
 
 If using the _filesystem_ storage type only, you can create the file in the project base directory:
```
fdevans, line 43: Let's add context here as well as to "why" they might consider this.  If it's a rare use case we can mention that as well.

Overall let's guide the user to picking the best path for managing this.  (Probably GUI?)

fdevans (changes requested):
Check out the comments.  I know you didn't change those sections, but let's address the page as a whole if we're going to make some updates

fdevans (approved):

PR #547, opened 2020-06-09
Update the OAuth2 SSO docs to reflect the changes made for 3.3.0

sjrd218:
Helps a user upgrade their OAuth setup to work with Rundeck 3.3.0.


fdevans (commented):
Do we need to leave some 3.2.x references since a there's a large base still on that?  Or do we include a link to the 3.2.x sub folder as a historical reference?

fdevans:
Going to approve and merge, then might add a few updates to cover the 3.2.x users that might need to reference how theirs should be configured.

PR #588, opened 2020-07-23
Job YAML export enhancement docs.

sjrd218:
Documents the auto line trimming feature for YAML exports.


fdevans:
Is this set as a default going forward?

sjrd218:
It's not the default, nor should it be. It enables customers who really want the nice looking export to be able to get it. Most people prefer the exact representation which is the default.

PR #589, opened 2020-07-23
Primary server id docs

sjrd218:
Primary server id docs


fdevans:
Does this need to be the same across all cluster instances?  I'm assuming based on placement it goes in rundeck-config.properties?

fdevans:
Merging for release, but may need additional updates based on question above

sjrd218:
Yes, the property should be set to the same server across all instances. The context of the placement of this property in the docs is documenting properties that are set in `rundeck-config.properties` so I think it's unnecessary to add that detail.

PR #591, opened 2020-08-04
Adding the documentation for HTTP Node Step

chrismcg14:
issue #583

PR #595, opened 2020-08-04
Update to workflow steps for http node request and log data values

chrismcg14:
issue #591


docs/manual/node-steps/builtin.md
```diff
@@ -135,6 +135,51 @@ Data
...
+
+Request Timeout
+: How long to wait for the request to complete before failing
+
+![HTTP Node Step - Authentication](~@assets/img/http-node-step-auth.png)
+
+#### Authentication 
+
+Authentication
+: Method of authentication to use
```
fdevans: Are there different options exist for the various Authentication methods?  Let's list the options they have here so it's covered and indexed by the doc search.

docs/manual/workflow-steps/builtin.md
```diff
@@ -127,6 +127,17 @@ Finally at the moment of run our job, if the referenced job defines in Job State
 
 ### Log Data Step
 
+The log data step allows you to log all the context data values and view them in a table. 
+
+![Log Data Step](~@assets/img/log-data-set.png)
+
+Debug Only
+: If checked, it will only log data at the debug level.
```
fdevans: I think this is more specifically "Only run if the job is run in Debug mode"?  Please confirm and clarify.

docs/manual/workflow-steps/builtin.md
```diff
@@ -127,6 +127,17 @@ Finally at the moment of run our job, if the referenced job defines in Job State
...
+
+After execution of the job, you will get a table like this:
+
+![Log Data Step Result](~@assets/img/log-data-result.png)
+
 ### Refresh Project Nodes
 
-### Data Step
```
fdevans, line 132: Was this intentionally removed?

PR #597, opened 2020-08-05
Adding the code for a new pager duty job steps page. Linked at the jo…

chrismcg14:
…b plugins page.


docs/manual/job-plugins.md
```diff
@@ -30,6 +30,7 @@
 - [Flow control](/manual/workflow-steps/builtin.md#flow-control)
 - [Job state conditional](/manual/workflow-steps/builtin.md#job-state-conditional)
 - [Log data step](/manual/workflow-steps/builtin.md#log-data-step)
+- [PagerDuty Steps (Enterprise)](/manual/workflow-steps/pagerduty.md#pager-duty-plugins-enterprise)
```
fdevans, line 33: Let's move this to the top of the list with the other Enterprise listings.

docs/manual/workflow-steps/pagerduty.md
```diff
@@ -0,0 +1,37 @@
+# Pager Duty Plugins (Enterprise)
```
fdevans: This looks like it only covers the Job Steps.  Let's also include the following on this page:
- Overview what PagerDuty is
- Distinguish between Job Steps and Notification Plugin options.
- Where do they get the API key?  (Note, they are different between the ones used for Job steps and ones used for Notification steps.)

Issue #507, opened 2020-05-20
Missing optional Node Attributes


pam-rundeck:
**Describe the bug**
Many of the optional node attributes are missing. it would be nice to have all predefined attributes in a single list

sshport - Connection port, default 22
ssh-keypath - Path to the SSH key in the Rundeck Key Store; for example keys/ec2/west.pem.
node-executor - 
file-copy-destination-dir
script-exec
script-exec-dir
script-exec-shell
file-copier
script-copy
script-copy-remote-filepath

**Source page**

This might not be the correct place but it is the largest consolidated list anywhere -
https://docs.rundeck.com/3.2.x/manual/document-format-reference/resource-yaml-v13.html#node-definition

PR #600, opened 2020-08-05
Updated the resources-yaml file to contain optional entries

chrismcg14:
Added the optional entries for node definitions


chrismcg14:
fixes #507

fdevans (approved):

PR #604, opened 2020-08-06
Updated the pictures on the page and added information about the vari…

chrismcg14:
…ous option for importing/exporting


fdevans:
Original comment wasn't accurate.  Only request is to add "some parts might be Enterprise"

fdevans (changes requested):
see conversation comment

fdevans (approved):

PR #611, opened 2020-08-10
added a link to instructions for installing and configuring windows r…

### LINKED ISSUE #264: Documentation bug in file administration/projects/node-execution/powershell.md
**Describe the bug**
Not clear if the following configuration is needed for CredSSP:
```
On the remote nodes
winrm quickconfig
​winrm set winrm/config/service/Auth @{Basic="true"}
winrm set winrm/config/service @{AllowUnencrypted="true"}
winrm set winrm/config/winrs @{MaxMemoryPerShellMB="1024"}
```

**Source page**
https://docs.rundeck.com/docs/administration/projects/node-execution/powershell.html#winrm-setting-to-use-powershell-plugin

chrismcg14:
…emote nodes. fixes #264

### DIFF
--- a/docs/administration/projects/node-execution/powershell.md
+++ b/docs/administration/projects/node-execution/powershell.md
@@ -80,7 +80,7 @@ keys/users/${job.username}.password
 
 ## WinRM Setting to use PowerShell Plugin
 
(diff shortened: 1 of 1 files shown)


docs/administration/projects/node-execution/powershell.md
```diff
@@ -186,6 +192,13 @@ To enable the CredSSP:
 winrm set winrm/config/service/auth @{CredSSP="true"}
 ```
 
+To enable unencrypted access run this command on the Rundeck server and the remote nodes:
```
fdevans: Why would someone want to do this?  Running "unencrypted" seems like a bad idea, so let's put context around the use case.  That way it doesn't seem like we're encouraging lower security.

Issue #529, opened 2020-06-02
Explanation needed for plugin dependencies


g3nsvrv:
**Describe the bug**
There's no mention in our documentation about install plugin dependencies on Rundeck; you can install them under $RDECK_BASE/server/lib or /var/lib/rundeck/lib depending on the installation method.

**Source page**
https://docs.rundeck.com/docs/developer/01-plugin-development.html#build-dependencies

g3nsvrv:
e.g. https://github.com/inokappa/rundeck-fluentd-plugin

gschueler:
If you need java dependencies it’s recommended to build a jar and include the dependencies inside the jar as documented under java plugin development.  It may also be possible to use groovy `@Grab` mechanism but I haven’t tried it

g3nsvrv:
@gschueler probably is a good practice to precompile the dependencies with the plugin but this thing actually works, the dependencies can be uploaded on the /var/lib/rundeck/lib folder.

pwonson:
Can we fix this page? 

https://docs.rundeck.com/docs/developer/01-plugin-development.html#java-plugin-development 

The correct path should be in this paragraph "If your Java classes require external libraries that are not included with the Rundeck runtime, you can include them in your .jar archive. (Look in $RDECK_BASE/tools/lib to see the set of third-party jars that are available for your classes by default at runtime)."

Replace "$RDECK_BASE/tools/lib" with  "/var/lib/rundeck/lib"

fdevans:
@chrismcg14 Let's fix @pwonson comment above.  The prior discussion will require more decisions on how to document.

g3nsvrv:
> Can we fix this page?
> 
> https://docs.rundeck.com/docs/developer/01-plugin-development.html#java-plugin-development
> 
> The correct path should be in this paragraph "If your Java classes require external libraries that are not included with the Rundeck runtime, you can include them in your .jar archive. (Look in $RDECK_BASE/tools/lib to see the set of third-party jars that are available for your classes by default at runtime)."
> 
> Replace "$RDECK_BASE/tools/lib" with "/var/lib/rundeck/lib"

@pwonson No need to replace, we need to included both methods, RDECK_BASE/tools/lib is for WAR, TOMCAT, and /var/lib/rundeck/lib is for RPM, DEB.

PR #629, opened 2020-08-27
fixed the incorrect path to third party jars

chrismcg14:
I did this because of issue #529 but I don't think this is a full fix.


g3nsvrv (commented):
should be both, RDECK_BASE is for TOMCAT and WAR deployments (windows included), and the /var/lib/rundeck is for RPM and DEB packages.

PR #635, opened 2020-08-28
Updated the API docs to v36 and added new endpoints.

chrismcg14:
Related to rundeck/rundeck#6299


ProTip:
Ideally this should be documented as an updated response with a new query parameter. Updated response particularly since we have broken from the previous API contract; the default response status is now `200` instead of `503`.

gschueler:
seems like duplicated work from @sjrd218  : https://github.com/rundeck/docs/pull/634

fdevans:
Closed as duplicate

PR #649, opened 2020-09-08
Update sso.md

g3nsvrv:
modify Rundeck autoconfigurl for better understanding of which url take from ping.


sjrd218 (changes requested):
I don't think this change is needed. When I created a Ping test account and set it up to connect to the auto config url, the one that is currently in the docs was correct. Your test account was created in a different way, so the url was slightly different.

Perhaps we should describe better that whatever url is used in the `autoConfigUrl` is a url that will automatically have `/.well-known/openid-configuration` appended to it, and the data that comes back from that url will be used to configure the SSO settings. If you can successfully pull the url that is made by `autoConfigUrl`+`/.well-known/openid-configuration` in your browser, then the auto configuration should work correctly.

fdevans:
closing per Stephen's comments

PR #670, opened 2020-09-25
add notification thread options

ronaveva:
documentation for PR https://github.com/rundeck/rundeck/pull/6494

### DIFF
--- a/docs/developer/05-notification-plugins.md
+++ b/docs/developer/05-notification-plugins.md
@@ -31,12 +31,12 @@ and passed in two sets of Map data:
 
 The Configuration data is fully custom depending on your plugin, and is described in the [Plugin configuration properties](#plugin-configuration-properties) section.
 
-## Notifications thread
+## Notifications thread feature
 
 Notifications can be run on their own thread by setting the following properties in the rundeck-config.properties file:
 
-`rundeck.notifications.own.thread`: ('true','false') When set to 'true' it will trigger the notifications on a separate thread.
-`rundeck.notifications.timeout`: (numeric, in miliseconds, defaults to 120000ms) When you set the notifications to be triggered on a separate thread, you can also set a time out for the notification thread so that it must be sent before a certain time.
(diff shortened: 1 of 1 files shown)


docs/developer/05-notification-plugins.md
```diff
@@ -31,6 +31,14 @@ and passed in two sets of Map data:
 
 The Configuration data is fully custom depending on your plugin, and is described in the [Plugin configuration properties](#plugin-configuration-properties) section.
 
+## Notifications thread
+
+Notifications can be run on their own thread by setting the following properties in the rundeck-config.properties file:
+
+`rundeck.config.notifications.own.thread`: ('true','false') When set to 'true' it will trigger the notifications on a separate thread.
+`rundeck.config.notifications.timeout`: (numeric, in miliseconds) When you set the notifications to be triggered on a separate thread, you can also set a time out for the notification thread so that it must be sent before a certain time.
```
fdevans: Is there a suggested value?  Is there a default value if this isn't specified?

fdevans (changes requested):
Need more details on timeout value

fdevans:
Also, is this something that's part of the current release (3.3.4) or is this related to a PR for 3.3.5?

ronaveva:
> 
> 
> Also, is this something that's part of the current release (3.3.4) or is this related to a PR for 3.3.5?

For 3.3.5

docs/developer/05-notification-plugins.md
ronaveva: There is no default time out, i'll add that to the doc.

fdevans: So it would never timeout?  It would go indefinitely?

ronaveva: Yes it would, similary to the current behavior. The only difference is that it would not stuck the job execution. You think it would be better to have a default time out? If that's so, what you think it would be an appropiate value for it?

fdevans (approved):

PR #676, opened 2020-10-02
Fix confusing hierarchy of Property Rendering Options

marcjay:
The current rendering of these options made me think that `valueConversionFailure`, `staticTextContentType` etc. were valid values for the `valueConversion` key
I believe the changes reflect the options as intended

Current view:
image


chrismcg14 (commented):
This looks good to merge @fdevans

fdevans:
Thanks for the update @marcjay !

PR #679, opened 2020-10-07
Created a new page describing the new notifications interface, added …

chrismcg14:
…warnings for how to turn it off on creating a job and job plugins. Linked the instructions on those pages.


docs/manual/notifications/new-interface.md
```diff
@@ -0,0 +1,13 @@
+# Notification Interface V2 (After v3.3.4)
```
fdevans: We don't typically mention version numbers in this context.  We just present it as the current interface so the title would be "Notification Interface" or something that describes this product area.

docs/manual/job-plugins.md
```diff
@@ -49,6 +49,16 @@ Notification plugins allow Rundeck to communicate changes in job execution state
 - [Email](/manual/notifications/email.md)
 - [Webhooks](/manual/notifications/webhooks.md)
 
+:::warning
```
fdevans: Not convinced these are necessary here.  Maybe just documenting it on the new page you created below that they can revert to the legacy interface but may run into issues.  This new UI was developed due to a scaling bug and is the long term path forward.

fdevans (changes requested):

docs/manual/notifications/new-interface.md
chrismcg14: Updated!

docs/manual/job-plugins.md
chrismcg14: Updated! @fdevans

PR #713, opened 2020-12-07
Fix of invalid URL @ oracle.md

n3y:
OLD : https://rundeck.org/docs/administration/configuration/configuration-file-reference.html#configuration-layout
NEW : https://docs.rundeck.com/docs/administration/configuration/config-file-reference.html#configuration-layout


fdevans:
Thank you @n3y

PR #714, opened 2020-12-09
Update upgrading.md

n3y:
Added a note that 'server.servlet.context-path' replaced 'server.contextPath' since version 3.3.

### DIFF
--- a/docs/upgrading/upgrading.md
+++ b/docs/upgrading/upgrading.md
@@ -130,4 +130,4 @@ mv rundeck-{{{rundeckVersionFull}}}.war $tomcat.base/webapps/rundeck.war
 
 - Due to changes in authentication, `tomcat-users.xml` and other Tomcat's authentication modules no longer work, you should configure users as described in [Authenticating Users](/administration/security/authentication.md#authenticating-users).
 - If you do not have `-Drundeck.config.location` defined or configured in `$tomcat.base/bin/setenv.sh` file (`tomcat.base\bin\setenv.bat` for Windows), Rundeck will read its config file from this location: `$rdeck.base/server/config/rundeck-config.properties`.
-- You **must** define the `server.contextPath` value in `rundeck-config.properties` to properly tell Rundeck about the context path used by tomcat. See [Installation on Tomcat](/administration/install/tomcat.md).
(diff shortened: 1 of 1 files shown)


fdevans:
@sjrd218 I like what was done here, but I think we usually flip this and include the current value by default.  Can you confirm that's the correct value for 3.3.x?

sjrd218:
`server.servlet.context-path` is the new config prop to use.

docs/upgrading/upgrading.md
```diff
@@ -130,4 +130,4 @@ mv rundeck-{{{rundeckVersionFull}}}.war $tomcat.base/webapps/rundeck.war
 
 - Due to changes in authentication, `tomcat-users.xml` and other Tomcat's authentication modules no longer work, you should configure users as described in [Authenticating Users](/administration/security/authentication.md#authenticating-users).
 - If you do not have `-Drundeck.config.location` defined or configured in `$tomcat.base/bin/setenv.sh` file (`tomcat.base\bin\setenv.bat` for Windows), Rundeck will read its config file from this location: `$rdeck.base/server/config/rundeck-config.properties`.
-- You **must** define the `server.contextPath` value in `rundeck-config.properties` to properly tell Rundeck about the context path used by tomcat. See [Installation on Tomcat](/administration/install/tomcat.md).
+- You **must** define the `server.contextPath` (`server.servlet.context-path` since version 3.3) value in `rundeck-config.properties` to properly tell Rundeck about the context path used by tomcat. See [Installation on Tomcat](/administration/install/tomcat.md).
```
fdevans, line 133: We generally like to include the current setting as the primary option and note the previous versions.  If you can change it to read as below I can approve it.   Thanks for the help!

```
You **must** define the `server.context-path` (`server.servlet.contextPath` for versions prior to 3.3) value in `rundeck-config.properties` to properly tell Rundeck about the context path used by tomcat. See [Installation on Tomcat](/administration/install/tomcat.md).
```

fdevans (changes requested):

docs/upgrading/upgrading.md
n3y: I agree with your proposal and will open a new pull request with the fixed/better version.
(Please note though that your comment contains wrong names of properties,
it's actually 'server.servlet.context-path' and 'server.contextPath'.)

PR #732, opened 2021-01-08
uploading thycotic documentation

chrismcg14:


docs/administration/security/storage-plugins/thycotic-storage.md
```diff
@@ -0,0 +1,37 @@
...
+: The username for a user in the account where the secrets should be stored.
+
+- **Password**
+: The password for a user in the account where the secrets should be stored.
+
+- **Address**
+: The base URL for the secret server account where the secrets should be saved. For example, `https://example.secretservercloud.com`.
+
+- **allowSelfSignedCert**
+: Set to true if wishing to use self-signed certificates.
```
fdevans: I think this needs a bit more context.  Allow them for auth to Thycotic or something else?

docs/administration/security/storage-plugins/thycotic-storage.md
```diff
@@ -0,0 +1,37 @@
...
+: This specifies the storage plugin to use. For Thycotic, it should always be "thycotic-storage."
+
+- **Path**
+: The path in Rundeck storage tree to apply the plugin. If just "keys" is specified, then all keys and directories added to Rundeck Key Storage will also be added to Thycotic. 
+
+- **removePathPrefix**
+: By default, the storage plugin will be invoked using the full path that is requested. If set to true, the path used when invoking the storage plugin would not include the prefix. It is recommended to set it to true. If set to false, keys will not be displayed unless a directory is specified in Rundeck.
+
+- **Username**
+: The username for a user in the account where the secrets should be stored.
```
fdevans: This should probably be something more like
UserName for an account with access to the secrets that will be used in Rundeck. 

Also should mention something along the lines of "it should have access to read/write secrets as needed" and also note it will be used by everyone that can manage secrets in Thycotic.

fdevans (changes requested):
I think we need a few more Use Case situations explained here. "Access all your Thycotic secrets from Rundeck to streamline and manage access"

fdevans (approved):

PR #738, opened 2021-01-22
add ingore all branches filter

travisrd:
add ingore all branches filter


ProTip (changes requested):
I believe we need to restore the branch filter for `build`:
```
branches:	
              only: /3\.[1-9]\..*/
```

The build should only be running on those branches.

travisrd:
Updated build workflow with 3.x branch filter.

ProTip (approved):

PR #747, opened 2021-02-04
Fix GitHub links in 3.3.9 Release Notes

marcjay:
Fix links to GitHub issues to use the UI URL instead of the API URL


fdevans:
Thanks Marc.  New generation process for that file. We'll fix that up for the next round.

PR #771, opened 2021-03-02
Update docker.md

### LINKED ISSUE #770: Missing information on LDAP example
**Describe the bug**
On the section **JAAS** it only references that, if we want to use LDAP configuration, we must call it them using the following format **RUNDECK_JAAS_LDAP_*** replacing the star with the name of the value that we want to include, e.g. **RUNDECK_JAAS_LDAP_USERIDATTRIBUTE**, but, this does not explain how to configure if the LDAP module will be required, sufficient, etc, (values explained here https://docs.rundeck.com/docs/administration/security/authentication.html#multiple-authentication-modules).

**Source page**
https://docs.rundeck.com/docs/administration/configuration/docker.html#user-authentication

jgarces-pd:
adding missing reference for **RUNDECK_JAAS_LDAP_FLAG**
should fix: https://github.com/rundeck/docs/issues/770
needs confirmation from @ltamaster

### DIFF
--- a/docs/administration/configuration/docker.md
+++ b/docs/administration/configuration/docker.md
@@ -71,6 +71,12 @@ RUNDECK_JAAS_FILE_*
 By convention the module name matches the name in the docs, and the config keys match
 the config options listed in the docs uppercase, and all one word.
 
+To configure the LDAP module as sufficient, required, requisite, optional, you can use:
+```
+RUNDECK_JAAS_LDAP_FLAG
+```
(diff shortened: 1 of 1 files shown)


jgarces-pd:
@jplassnibatt confirmed that is on:
https://github.com/rundeck/rundeck/blob/eb86dc2d6d56ec434a53d0ddaeaac915bb53a1ce/docker/official/remco/templates/jaas-loginmodule.conf#L4

@fdevans please kindly approve this to include this reference in our doc.

PR #781, opened 2021-03-15
API v39 update

chrismcg14:
updated the docs to refelct api version 39, min version of 11 in 3.4.0 and noted removed endpoints and their replacements.

### DIFF
--- a/docs/api/rundeck-api.md
+++ b/docs/api/rundeck-api.md
@@ -25,8 +25,8 @@ Minimum
 For API endpoints described in this document, the *minimum* API version required for their
 use is indicated by the URL used, e.g.:
 
-    /api/2/system/info
-    /api/1/projects
+    /api/11/system/info
+    /api/14/projects
 
 This means you must use at least the API version indicated to access the
 endpoint, unless otherwise noted. Some features or functionality for the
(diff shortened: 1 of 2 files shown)


docs/api/rundeck-api.md
```diff
@@ -81,6 +78,22 @@ Changes introduced by API Version number:
 **Deprecation**
 * API versions below `{{{ apiDepVersion }}}` are *deprecated*.  Clients using earlier versions should upgrade to use `{{{ apiDepVersion }}}` as the minimum version before release `{{{ apiDepRelease }}}` to avoid errors.
```
gschueler: could remove the text about deprecated versions now i think

chrismcg14:
@gschueler I removed the wrong deprecation text the first time I believe. Should be good now.

gschueler:
@chrismcg14 there is also this text https://github.com/rundeck/docs/pull/781/files#diff-5547774df67712f214d7ebe2259fbbdb03d59742d33a8768f5ffb77073e44ac1R73

chrismcg14:
@gschueler yeah I wasn't sure which deprecation text you meant, they are both removed now

PR #790, opened 2021-04-01
added info for configuring v3 key for PagerDuty web hook

chrismcg14:
Updated docs for using V3 with the private key.


fdevans:
Since this is a short term solution I'd like to leave this out of official docs until we support the Key Storage method of storing the private key.

PR #809, opened 2021-04-27
3.3 prep Learning Section, Tabs, and Migration content

fdevans:
Establishes new Learning menu, adds tabs plugin to 3.3.

Install How-To might need latest commands @ProTip is fixing


fdevans:
implemented by protip another way

Issue #819, opened 2021-05-07
logrotate rule with some issues with deb installer.


luisnunezsalgado:
**Describe the bug**
We need to clarify on this documentation that we need to define the owner/group of the file to be rotated (service.log). If you install the deb package and using it, the owner of the service.log is rundeck/adm. With that in mind, the documentation doesn't mention anything about this and defining the logrotate rule as following. 
 
-------------
Create logrotate configuration file
```

 cat << EOF > /etc/logrotate.d/theRundeckLog
 /var/log/rundeck/service.log {
  su root root
...
  create 640 root adm
```
-------------
Causing that the service.log doesn't be able to receive new information on it. Because the file remains with the root/root owner. 

we need to clarify that we need to define in those lines the owner/group of the service.log file, in order to avoid any further error. 

case related: 
00038266 - https://rundeck.lightning.force.com/lightning/r/Case/5006e00001RpJh3AAF/view

**Source page**
https://docs.rundeck.com/docs/administration/maintenance/logs.html

Thanks in advance. 

Regards, 
Luis

fdevans:
Thanks for the report. As part of a cleanup of our documentation issue tracker, we're closing issues older than 12 months — the docs site has been substantially reorganized and rewritten since older reports were filed, so we can't be confident this still applies to the current content. This isn't a judgment that the report was invalid.

If this is still an issue on the current docs site (docs.rundeck.com), please open a new issue linking the specific page and describing what's wrong. Please don't reopen this issue — a fresh issue keeps things properly tracked against the current site.

Thanks for your patience,
The Rundeck Team

PR #820, opened 2021-05-07
Update logs.md

luisnunezsalgado:
Fixes: https://github.com/rundeck/docs/issues/819


fdevans:
I don't think this makes the necessary changes to fix #819 very clear.

fdevans (changes requested):
This needs a bit more detail to address the issues in #819.

PR #826, opened 2021-05-17
Include docs for refresh healthCheck cache feature

carlosrfranco:
Include documentation for the refresh health status cache feature added to fix the issue https://github.com/rundeckpro/rundeckpro/issues/1539


docs/manual/healthchecks.md
```diff
@@ -73,6 +73,8 @@ Each Health Check will result in a Health Status:
 	```properties
 	project.healthcheck.enabled=true
 	```
+ 
+    The health check uses a cache to store the statuses and improve performance when requesting them. Then, enable the "Refresh health status cache" to clean and update the cache periodically and set the update period in the "Cache refresh period" field whose default value is 30 seconds.
```
gschueler:
```suggestion
    The health check uses a cache to store the statuses and improve performance when requesting them. To automatically refresh the Health Checks, enable the "Refresh health status cache" and set the update period in the "Cache refresh period" field whose default value is 30 seconds. 
```

gschueler (changes requested):

gschueler (approved):

PR #827, opened 2021-05-17
Update Logo to Rundeck by PagerDuty

mishingo:
- add additional CSS to increase the size of the logo


fdevans:
This PR should be against 3.4.x (default) branch and we can port back to 3.3.x early if needed.

PR #837, opened 2021-05-20
added docs to show we now support bcrypt type password hashes

chrismcg14:


docs/administration/security/authentication.md
```diff
@@ -51,6 +51,10 @@ The list of roles can be accepted as-is (default), or you can add a prefix to th
 
 ## PropertyFileLoginModule
 
+:::warning
+It is reccomended to use BCRYPT versus other methods, this will lead to the best results.
```
ProTip:
```suggestion
It is recommended to use `BCRYPT` encrypted passwords with `realm.properties` as it is the most secure option available. Avoid `plain`, `MD5`, and `CRYPT`.
```

Adds more context and stronger language

docs/administration/security/authentication.md
```diff
@@ -110,11 +114,12 @@ mypass    <-----Type this value
 obfuscate: OBF:1xfd1zt11uha1ugg1zsp1xfp
 md5: MD5:a029d0df84eb5549c641e04a9ef389e5
 crypt: CRYPT:jsnDAc2Xk4W4o
+bcrypt: BCRYPT:JDG45dJSNnsf4
```
ProTip: This doesn't look like a `bcrypt` hash ;)

ProTip (changes requested):
It may be worth discussing/mentioning the caveat with the salt revisions the PR talks about. In case somebody tries using `bcrypt` hashes generated from different tools.

ProTip:
> In case somebody tries using `bcrypt` hashes generated from different tools.

Which is how it was discovered haha.

PR #869, opened 2021-08-02
Update email-settings.md

### LINKED ISSUE #856: Email Groovy and Properties config are wrong
## **Describe the bug**
Email Groovy and Properties config are wrong. On RPM and WAR installations you can use both without "translating the rundeck-config.properties into a groovy format", and is being detected by Rundeck automatically without making any extra configuration, like mentioned here that it says is required (not required at all):

https://docs.rundeck.com/docs/administration/configuration/config-file-reference.html#specify-config-file-location

Also this is wrong:

```
Specify config file location
You will need to point rundeck at the new filename when you start up rundeck:

Launcher:

  java -jar -Drundeck.config.name=rundeck-config.groovy rundeck-launcher.jar
Copied!
RPM: Add this to the /etc/sysconfig/rundeckd file:

    export RDECK_CONFIG_FILE="/etc/rundeck/rundeck-config.groovy"
Copied!
RPM/DEB: Add this to the /etc/default/rundeckd file:

    export RDECK_CONFIG_FILE="/etc/rundeck/rundeck-config.groovy"
```

Based on this you can configure both files.

```
Advanced SMTP settings
If you need more advanced configuration (e.g., authenticated and secured over SSL), see the grails Mail plugin configuration: Grails Mail Configuration(opens new window)

The caveat for using this is that in our experience it requires using a .groovy formatted configuration file to support the extended configuration properties.

See Groovy config format.
```

From: https://docs.rundeck.com/docs/administration/configuration/config-file-reference.html#groovy-config-format
```
Groovy config format
If you would prefer to use Groovy for the config file, you can use rundeck-config.groovy instead of rundeck-config.properties. Or, you can use a combination of the two (i.e. some settings configured in the properties file and some in the Groovy file).
```

This was tested on 3.3.11 RPM and 3.3.8 WAR windows and works perfect.

## **Source page**
https://docs.rundeck.com/docs/administration/configuration/email-settings.html#advanced-smtp-settings

jgarces-pd:
fixes: https://github.com/rundeck/docs/issues/856

Change on the Line 35 to clarify that is not required to translate every config from properties to groovy format.

### DIFF
--- a/docs/administration/configuration/config-file-reference.md
+++ b/docs/administration/configuration/config-file-reference.md
@@ -676,18 +676,27 @@ a{
 }
 ```
 
-### Specify config file location
+### Specify config file location (optional)
+:: warning
+This configuration should only be applied if the location or format of the rundeck-config is different from default. e.g. groovy format
+::
 
 You will need to point rundeck at the new filename when you start up rundeck:
 
 - Launcher:
+```bash
+java -jar -Drundeck.config.name=rundeck-config.groovy rundeck-launcher.jar
+```
+or
+```bash
+java -jar -Drundeck.config.name=/opt/rundeck/config/rundeck-config.properties rundeck-launcher.jar
+```
 
-        java -jar -Drundeck.config.name=rundeck-config.groovy rundeck-launcher.jar
-
-RPM: Add this to the `/etc/sysconfig/rundeckd` file:
-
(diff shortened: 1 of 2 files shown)


fdevans:
this is now supported in Config Management and .groovy is no longer needed.

PR #876, opened 2021-09-01
Adding doc for new property setting rundeck.gui.matchedNodesMaxCount

carlosrfranco:
The property `rundeck.gui.matchedNodesMaxCount` was included to define maximum number of nodes to be shown on Matched Nodes section

PR: https://github.com/rundeck/rundeck/pull/7215


fdevans:
We should also add an entry to /administration/configuration/gui-customization.md if it is not there already.

fdevans (changes requested):
add details to GUI Customization page.

fdevans (approved):

PR #882, opened 2021-09-29
RUN-409: Document config management API

qualman:
First pass at documenting the config management API. Keeping as draft until I get a review.

### DIFF
--- a/docs/api/rundeck-api.md
+++ b/docs/api/rundeck-api.md
@@ -1075,7 +1075,7 @@ Manage your plugin and custom configuration properties via API.
 
 ### Create or Update Configurations ####
     POST api/36/config/save
-Create or update configs and properties. Invalid config data is ignored.
+Create or update configs and properties. POST fails if any configs are invalid.
 
 **Request:**
 
@@ -1107,9 +1107,6 @@ Example JSON
     "key": "myNewCustomConfig",
     "value": "valueOfNewCustomConfig",
     "strata": "default"
-  },
-  {
-    "key": "myIncompleteData"
   }
 ]
 ```
@@ -1118,7 +1115,7 @@ Example JSON
 
     200 OK
 
-One or more configs were successfully saved or updated. A payload reflecting save or creation status is returned.
+All configs were successfully saved or updated. A payload reflecting save or creation status is returned.
 
 Headers  
 `Content-Type: application/json`:
@@ -1126,7 +1123,7 @@ Headers
 Body
 ```json
 {
-  "msg":"Saved, with some configs ignored",
+  "msg":"Saved",
   "created": [
     {
       "key": "myNewCustomConfig",
@@ -1140,18 +1137,13 @@ Body
       "value": "newValueForCustomConfig",
       "strata": "default"
     }
-  ],
-  "ignored": [
-    {
-      "key": "myIncompleteData"
-    }
-  ],
+  ]
 }
 ```
 ---
     400 Bad Request
 
-All configs were formatted in an invalid way and were ignored. Check your request data and try again.
+Some or all configs were formatted in an invalid way and were ignored. Check your request data and try again.
 
 ---
     401 Unathorized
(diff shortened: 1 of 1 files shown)


docs/api/rundeck-api.md
```diff
@@ -1069,14 +1069,161 @@ Response:
...
+
+`Content-Type: application/json`:
+
+```json
+{
+    "msg":"saved"
+}
+```
+
+### List configmeta.json configs ####
```
fdevans: I don't think we need to make this one public.  The configmeta.json was the current "quick" way to implement and we might change how we populate this list later.

docs/api/rundeck-api.md
```diff
@@ -1069,14 +1069,161 @@ Response:
...
+  }
+```
+
+**Response**
+
+`Content-Type: application/json`:
+
+``` json
+{
+    "msg": "Deleted!!"
```
fdevans: Not Docs related, but could we get a bit more information in this output?  What was deleted, etc.

Also... we don't need two exclamation points in there.

docs/api/rundeck-api.md
```diff
@@ -1069,14 +1069,161 @@ Response:
...
+}
+```
+
+**Response**
+
+`Content-Type: application/json`:
+
+```json
+{
+    "msg":"saved"
```
fdevans: Not Docs related, but it would be nice to have more information back on success calls.  What was updated, etc.

Also what happens if one entry fails?  Does the whole payload fail?

fdevans (changes requested):

docs/api/rundeck-api.md
qualman: Deleted!

qualman:
The latest PR fully documents existing error/unsuccessful responses. At this point, we are going to make a few optimizations to the API endpoints themselves, and this PR will reflect them just-ahead-of-time.

docs/api/rundeck-api.md
```diff
@@ -1069,24 +1069,318 @@ Response:
...
 
+Headers
+* `Content-Type` Optional
+  * Accepts: `application/json`
+
+**Responses:**
+
+    200 OK
+
+Call was successful. A list of all configs is returned.
```
fdevans: I think we need to clarify that this is all configs stored using Configuration Management.  For a list of "All Configs" reference System Report.

docs/api/rundeck-api.md
```diff
@@ -1069,24 +1069,318 @@ Response:
...
+
+Headers
+* `Content-Type` Optional
+  * Accepts: `application/json`
+
 **Response**
 
+    200 OK
+
+The call was successful. Rundeck will reread the config properties file.
```
fdevans, line 1300: I think it's important to note here that some configs require a restart of Rundeck and this refresh will only work on certain configs.

fdevans (changes requested):

docs/api/rundeck-api.md
qualman: Added as a note

docs/api/rundeck-api.md
qualman: Added as a change/note

docs/api/rundeck-api.md
```diff
@@ -1069,24 +1069,323 @@ Response:
...
+        "strata": "Server",
+        "category": "Custom",
+        "visibility": "Custom"
+      },
+      {
+        "key": "myCustomConfig2",
+        "value": "myConfigValue2",
+        "strata": "default",
+        "category": "Custom",
+        "visibility": "Custom"
```
imad6639: Not what the api is like.  This is the editorData

docs/api/rundeck-api.md
```diff
@@ -1069,24 +1069,323 @@ Response:
...
+
+**Responses**
+
+    200 OK
+
+All configs were successfully saved or updated. A payload reflecting save or creation status is returned.
+
+Headers  
+`Content-Type: application/json`:
+
```
imad6639, line 1122: do you want to add in the other use cases as mentioned in the work request?

{
    "msg": "Saved, with some invalid configs ignored",
    "created": [],
    "updated": [
        "r.feature.notificationsOwnThread.enabled"
    ],
    "ignored": [
        "rundeck.feature.notificationsOwnThread.enabled"
    ]
}

qualman: I changed that in the ticket because there was some slack discussion that we wanted the whole thing to fail if any data was invalid

docs/api/rundeck-api.md
qualman: Just fixed it :)

fdevans (approved):

imad6639 (approved):

qualman:
I'm not allowed to merge on this repo. Can either of you @imad6639 @fdevans do that? TY!

PR #887, opened 2021-10-20
Fix google groups email address

gschueler:


fdevans:
I think this was a duplicate PR to #889 which I already merged.

PR #898, opened 2021-11-29
Initial CLI Docs

fdevans:
Implements the `rd` CLI tool docs in the main docs site with current content.


fdevans:
Going to merge to get some other changes out as well and start some marketing efforts around the learning articles.  if there are fixups later we can address quicly.

PR #901, opened 2021-12-05
Update ssh-on-linux-nodes.md

timhaak:
Document says leave file name in.

But no filename in the path given.


fdevans (approved):

fdevans:
Thank you @timhaak Good Catch!

PR #902, opened 2021-12-13
s/log4j/log4j2

nbarrientos:
Rather a question first, are you sure that `-Dlog4j.formatMsgNoLookups=true` mitigates the issue? All the advisories that I've seen mention `log4j2.formatMsgNoLookups ` instead, ex:

* https://access.redhat.com/security/cve/cve-2021-44228
* https://security-tracker.debian.org/tracker/CVE-2021-44228


gschueler:
thanks, it was corrected

PR #918, opened 2022-02-02
Add warning about comments in ldap configuration

sclarson:
When reconfiguring our rundeck ldap integration I restarted and started receiving 503 errors after login.

The root cause of this was found to be a comment in the ldap configuration.  When searching for more information we found documentation of this in the confluent docs: https://docs.confluent.io/5.4.5/control-center/security/c3-auth-ldap.html

This is probably not what the final wording should be, but it is enough to express the intent.

Software version this happened on was: 2.11.9-1


fdevans (approved):
@sclarson, Thanks for highlighting this for us!

PR #928, opened 2022-02-16
Documentation for Rundeck Plugin Blocklist

chrismcg14:


docs/administration/security/blocklist.md
```diff
@@ -0,0 +1,20 @@
+# Plugin Blocklist
+
+To use the plugin blocklist, you will need to create a blacklist yaml file in the following format:
```
fdevans: Let's change `blacklist` to `blocklist`

docs/administration/security/blocklist.md
```diff
@@ -0,0 +1,20 @@
+# Plugin Blocklist
```
fdevans, line 1: We probably need a bit more of a narrative on "why" someone would do this.

fdevans (changes requested):

fdevans:
@chrismcg14 I made some updates, but still need a bit more info:
- Where should the file be placed for each installation method?
- Is there a specific name for the file?

fdevans (approved):

PR #931, opened 2022-02-23
RUN-636: Added a note because this plugin present some problems with Windows Server nodes

alexander-variacode:
Issue:
https://github.com/rundeckpro/rundeckpro/issues/2097

Description:
Added a note at the end of the plugin documentation because this plugin could present some problems when it tries to copy a file to Windows Server nodes.


docs/administration/projects/node-execution/openssh.md
```diff
@@ -44,3 +44,11 @@ Also, you can define the configuration at Node Level, setting the node-executor
        ssh-password-storage-path ="keys/node/user.password"
        ssh-options="-o ConnectTimeout=5000"/>
 ```
+
```
fdevans: This doesn't make any sense.  Why do we need to specify that OpenSSH is required for OpenSSH plugins?  That should be implied.  Windows servers are not OpenSSH.

alexander-variacode: This is because Windows does not have OpenSSH natively, it has to be installed in Windows Server 2019 and it has had some problems when you try to copy a file from Linux to Windows using the command scp. For that reason, it would be a good idea to change that command to rsync or sftp. Another good reason to change scp is because it has a security problem, which is described on this ticket https://github.com/rundeckpro/rundeckpro/issues/2345.

fdevans (approved):

PR #932, opened 2022-02-23
RUN-595: documenting Log output limit optional property

perezo-pd:
Issue:
https://github.com/rundeckpro/rundeckpro/issues/1850

Description:
Documenting new optional property to limit the amount of logs being displayed

### DIFF
--- a/docs/administration/configuration/config-file-reference.md
+++ b/docs/administration/configuration/config-file-reference.md
@@ -653,7 +653,7 @@ First Load Asynch: true/false
 
 ### Limit displayed Job execution Log Output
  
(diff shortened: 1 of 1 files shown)


carlosrfranco (commented):
I think it would be good to explain on the doc that this property should limit the log output when following the execution that is running (not finished yet)

perezo-pd:
> I think it would be good to explain on the doc that this property should limit the log output when following the execution that is running (not finished yet)

Hi @carlosrfranco , I updated the doc with the detail you mentioned

fdevans (approved):

PR #956, opened 2022-04-19
Redo with new images

ddarby-hike:
I have no idea if this is the correct way to make the change or not....


fdevans:
Closing after working with Dave on some training and better access

PR #962, opened 2022-04-29
log4j -> log4j2 fix.

ddarby-hike:


fdevans:
`sudo` is implied if the user isn't installing as root. For this to work they would have to have `sudoers` setup, etc.  

In general we leave the commands as the base command needed and they can decide how to execute with the proper privileges.

Take that part off and I can approve the fix for Azure docs.

fdevans (approved):

ddarby-hike:
The Ubuntu/Debian install
<https://docs.rundeck.com/docs/administration/install/linux-deb.html> is
currently 50/50 (sudo/no sudo).  I was simply making them consistent by
adding sudo.  Do you want me to remove the sudo's on the page instead?

I've seen more customers use sudo, but I don't have a preference and just
prefer to be consistent.  So, I'll defer to you...

Thanks,
-David

PR #963, opened 2022-05-03
Added MySQL create user step

khudgins:
MySQL 8.x has deprecated the shortcut that allows you to create a new database user with a grant statement. In order to minimize confusion for newbies who are using MySQL, I've patched the detailed mysql doc to break out the create user step into a separate statement. This should work for all MySQL installs, regardless of version, so it should be more generically useful for everyone.


khudgins:
Standby, more commits incoming.

fdevans (approved):

PR #966, opened 2022-05-05
RUN-872: Execution log settings

alexander-variacode:
Add documentation regarding the fix from this issue RUN-138.
Related PR: https://github.com/rundeck/rundeck/pull/7248

### DIFF
--- a/docs/administration/configuration/config-file-reference.md
+++ b/docs/administration/configuration/config-file-reference.md
@@ -513,7 +513,7 @@ Delay is in milliseconds. If a max is set to `-1`, then retries will happen inde
 
 ### Execution log settings
 
-You can globally specify for all your jobs (regardless of which project they belong to) a log size limit and action as you needed. If you configure a job with other values, they override the global settings.
+You can globally specify for all your jobs (regardless of which project they belong to) a log size limit and action as you needed. If you configure a job with other values, the process will take the minimum value and prioritize Halt over Truncate.
 
 ```properties
 rundeck.execution.logs.output.limit = 5MB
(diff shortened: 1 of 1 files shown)


fdevans:
I think we need to explain this a bit more clearly.  Is this limit per job, per project or everywhere?  Halt and Truncate seem like the same thing depending on where it's applied.

docs/administration/configuration/config-file-reference.md
```diff
@@ -511,6 +511,17 @@ rundeck.execution.stats.retryDelay=5000
 
 Delay is in milliseconds. If a max is set to `-1`, then retries will happen indefinitely.
 
+### Execution log settings
+
+You can globally specify for all your jobs a log size limit and action as you needed.
```
fdevans: See other comment

fdevans (changes requested):

docs/administration/configuration/config-file-reference.md
```diff
@@ -511,6 +511,19 @@ rundeck.execution.stats.retryDelay=5000
...
+
+You can globally specify for all your jobs (regardless of which project they belong to) a log size limit and action as you needed. If you configure a job with other values, they override the global settings.
+
+```properties
+rundeck.execution.logs.output.limit = 5MB
+rundeck.execution.logs.output.limitAction = truncate
+```
+* `rundeck.execution.logs.output.limit`. Show an error when that limit is exceeded. Could be entered either maximum total line-count (e.g. "100"), maximum per-node line-count ("100/node"), or maximum log file size ("100MB", "100KB", etc.), using "GB","MB","KB","B" as Giga- Mega- Kilo- and bytes.
+* `rundeck.execution.logs.output.limitAction`. Is the action applied when the limit is achieved. It is triggered only if there is a value at rundeck.execution.logs.output.limit. The values accepted are ‘truncate' or 'halt’. The default value if it does not set is 'halt'.
+    * a. `halt`: the job will halt.
```
fdevans: The job itself will stop processing?

alexander-variacode: I do not know what would be the correct operation of this because in the current documentation of this (https://docs.rundeck.com/docs/manual/creating-jobs.html#log-limit) it says "Halt - the job will halt with a certain status" but when I test it the log output never stops growing. I am going to ask to the team and rewrite this feature if it is necessary.

alexander-variacode:
I fixed a problem on global log limit for a job and I updated the doc related to
Issue:
https://pagerduty.atlassian.net/browse/RUN-1013

PR #973, opened 2022-05-19
RUN-758: add docs for enable/disable public key GUI download

ltamaster:
Related to https://github.com/rundeck/rundeck/pull/7691

### DIFF
--- a/docs/administration/configuration/gui-customization.md
+++ b/docs/administration/configuration/gui-customization.md
@@ -242,8 +242,8 @@ Sets the maximum number of nodes to be displayed on [Matched Nodes](/manual/crea
 - Example: ```(Default: true)```
 - min version: 4.2.0
 
-Enable/disable the public key GUI download option in the Key Storage 
-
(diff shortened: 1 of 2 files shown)


docs/administration/configuration/gui-customization.md
```diff
@@ -238,6 +238,13 @@ Sets the default value for users logged status to show on summary page (it will
 Sets the maximum number of nodes to be displayed on [Matched Nodes](/manual/creating-jobs.md#node-dispatching-and-filtering) session in job edit page.
 
 
+### rundeck.gui.keystorage.downloadenabled
+- Example: ```(Default: true)```
+- min version: 4.2.0
+
+Enable/disable the public key GUI download option in the Key Storage
```
fdevans: Need to clarify here what "true" mean and what "false" means.  Does "True" mean I can download or can't download?

fdevans (changes requested):

fdevans (approved):

Issue #39, opened 2018-07-27
Deployment bug: old files that were removed are still present in published docs


gschueler:
e.g.

http://rundeck.org/3.0.x-SNAPSHOT/administration/configuration/storage/using-mysql-as-a-database-backend.html
http://rundeck.org/3.0.x-SNAPSHOT/administration/scaling/storage/mysql-setup-guide.html

these files are no longer generated from current docs source, however the website still has them.

old files should be removed from the s3 bucket when new site is deployed

PR #979, opened 2022-06-01
Docs for RUN-913 fix

carlosrfranco:


docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,29 @@
 # Rundeck 4.3 Upgrade Notes
 
 :::warning
-Any keys or passwords created using encryption in 4.2.0/4.2.1 will be lost when upgrading, it is recommended to back up your data before upgrading from those versions. Data created in earlier versions (before v4.2.0) is will transfer properly.
-:::
\ No newline at end of file
+Any keys or passwords created using encryption in 4.2.0/4.2.1 will be lost when upgrading, it is recommended to back up your data before upgrading from those versions. Data created in earlier versions (before v4.2.0) will transfer properly.
```
dileshni, line 4: Should we remove this line?

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,29 @@
...
+:::
+
+## Private keys and password encryption issue
+
+The release 4.3.0 fixes an issue that results in invalid credentials after upgrading from version 3.4.x to version 4.2.0.
+
+This issue is related to the key encryption method and causes keys already stored using previous versions to become invalid.
+
+One of the impacts related to this is related to the private keys used in SCM become invalid disabling all projects that use SCM with SSH private keys.
```
dileshni: Typos

`
This issue is related to the key encryption method and causes keys already stored using previous versions to become invalid.

Specifically, private keys used in SCM becoming invalid after the upgrade. This results in disabling all projects at that use SCM with SSH private keys.`

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,29 @@
...
+
+## Upgrading from 4.2.0
+
+:::tip
+**The issue was detected in version 4.2.0 and, if no action was taken so far, the stored keys should work normally in version 4.3.0.** 
+:::
+
+However, if the keys have been overwritten by the same key in order to workaround the issue in 4.2.0 version, the same step must be performed again, i.e., the keys must be overwritten by the same private key in version 4.3.0
```
dileshni: This sentence is hard to read. Can you reconsider re-wording it please?

docs/upgrading/upgrading-to-4.3.md
dileshni: consider rewording to

`An issue was reported in 4.2.0 and 4.2.1 where keys and passwords created using encryption was lost when upgrading. We recommend backing up your data before upgrading from these two versions. Data created prior to version 4.2.0 will transfer correctly.`

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,29 @@
...
+
+This issue is related to the key encryption method and causes keys already stored using previous versions to become invalid.
+
+One of the impacts related to this is related to the private keys used in SCM become invalid disabling all projects that use SCM with SSH private keys.
+
+## Upgrading from 4.2.0
+
+:::tip
+**The issue was detected in version 4.2.0 and, if no action was taken so far, the stored keys should work normally in version 4.3.0.** 
+:::
```
ehe-pd: Here is my suggestion:

This issue was detected after upgrading from v3.4.x to v4.2.x. 

- If you didn't change the stored private keys after the upgrading then you don't need to do anything in v4.3.0.
- If you have already overwritten private keys in v4.2.x to work around the issue, you have to overwrite those private keys again after upgrading to v4.3.0

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,29 @@
...
+
+However, if the keys have been overwritten by the same key in order to workaround the issue in 4.2.0 version, the same step must be performed again, i.e., the keys must be overwritten by the same private key in version 4.3.0
+
+Steps to overwrite key storage:
+
+- Go to key storage page
+- Select the private key
+- Click on Overwrite Key button
+- Overwrite the key **with the same key**
+- Check if the keys is working well
```
ehe-pd: Maybe we can leverage our online document to simplify those steps:

- Go to the key storage page
- Select the private key need to be overwritten then click the `Overwrite Key` button in the toolbar
- Follow the same steps as [Add a Private Key](https://docs.rundeck.com/docs/manual/system-configs.html#key-storage) to overwrite the keys

ehe-pd (changes requested):

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,29 @@
 # Rundeck 4.3 Upgrade Notes
 
 :::warning
-Any keys or passwords created using encryption in 4.2.0/4.2.1 will be lost when upgrading, it is recommended to back up your data before upgrading from those versions. Data created in earlier versions (before v4.2.0) is will transfer properly.
-:::
\ No newline at end of file
+An issue was reported in 4.2.0 and 4.2.1 where keys and passwords created using encryption was lost when upgrading. We recommend backing up your data before upgrading from these two versions. Data created prior to version 4.2.0 will transfer correctly.
```
qualman: Recommending even more clarity on this line:

> When upgrading from 4.2.0 and 4.2.1, some users have reported keys and passwords created using encryption were lost. We **strongly recommend** backing up your data before upgrading from these two versions. Data created prior to version 4.2.0 should not be affected.

gschueler: Suggest also clarifying: 
```suggestion
An issue was reported in 4.2.0 and 4.2.1 where keys and passwords created using encryption was lost when upgrading from version 4.1.0 and earlier. We recommend backing up your data before upgrading from these two versions. Data created prior to version 4.2.0 will transfer correctly.
```

qualman: Oh, how about this then, to combine them both:

> When upgrading from 4.1 to 4.2, some users reported keys and passwords created using encryption were lost. Therefore, we **strongly recommend** backing up your data before upgrading from 4.2 to 4.3. Data created _prior to version 4.2_ should not be affected when upgrading to 4.3.

docs/upgrading/upgrading-to-4.3.md
gschueler: This was also present upgrading from 4.1, correct?

qualman:
One thing I am struggling to understand overall is whether or not these issues are supposed to be related. As a user who doesn't have any context for these issues, all of their titles involve issues with private keys. Perhaps we edit the titles to be more specific to the issues, or specifically note that they are wholly separate issues from one another.

docs/upgrading/upgrading-to-4.3.md
dileshni: Yes, it was present upgrading from 4.1 as well

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,27 @@
 # Rundeck 4.3 Upgrade Notes
 
 :::warning
-Any keys or passwords created using encryption in 4.2.0/4.2.1 will be lost when upgrading, it is recommended to back up your data before upgrading from those versions. Data created in earlier versions (before v4.2.0) is will transfer properly.
-:::
\ No newline at end of file
+When upgrading from 4.1 to 4.2, some users reported keys and passwords created using encryption were lost. Therefore, we strongly recommend backing up your data before upgrading from 4.2 to 4.3. Data created prior to version 4.2 should not be affected when upgrading to 4.3.
```
fdevans: What was the resolution on previous versions?  I think I read that it doesn't matter what version someone came from.  It was just the fact that they were using 4.2 right?

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,27 @@
...
+:::
+
+## Private keys and password encryption issue
+
+The release 4.3.0 fixes an issue that results in invalid credentials after upgrading from version 3.4.x to version 4.2.0.
```
fdevans: This sentence conflicts with the warning above.

maybe it needs to be "Upgrading from any previous version to 4.2.0"

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,27 @@
...
+
+This issue is related to the key encryption method and causes keys already stored using previous versions to become invalid.
```
fdevans: This sentence is confusing.  Can we articulate it in a way that uses an example  of what a customer might have done?

ehe-pd:
One more amendment for the SSH Key issue specifically related to GitHub users:



For Github SCM users: 
[Github has stopped supporting insecure algorithms](https://github.blog/2021-09-01-improving-git-protocol-security-github/#when-are-these-changes-effective), and our current SSH implementation doesn't support the algorithms required by Github so you cannot use `Private Key` method to authenticate Rundeck client to Github. As a workaround, You still can use the `Password` authentication method with the Github `Personal Access Token` to connect to your Github repos. 

We are working on upgrading the SSH implementation to solve this issue in a future release.

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,27 @@
...
+
+Specifically, private keys used in SCM becoming invalid after the upgrade. This results in disabling all projects at that use SCM with SSH private keys.
+
+## Upgrading from 4.2.0
+
+This issue was detected after upgrading to v4.2.x from any previous version.
+
+- If you didn't change the stored private keys after the upgrading then you don't need to do anything in v4.3.0.
```
fdevans: Is it only "change" an existing key?   Or is it "If you didn't add/change any keys stored in the Key Storage..."?

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,27 @@
...
+
+## Private keys and password encryption issue
+
+The release 4.3.0 fixes an issue that results in invalid credentials after upgrading from version 3.4.x to version 4.2.0.
+
+This issue is related to the key encryption method and causes keys already stored using previous versions to become invalid.
+
+Specifically, private keys used in SCM becoming invalid after the upgrade. This results in disabling all projects at that use SCM with SSH private keys.
+
+## Upgrading from 4.2.0
```
fdevans: What if the customer is using a Key Storage plugin like CyberArk, Thycotic, Vault, etc.?

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,27 @@
...
+
+This issue was detected after upgrading to v4.2.x from any previous version.
+
+- If you didn't change the stored private keys after the upgrading then you don't need to do anything in v4.3.0.
+
+- If you have already overwritten private keys in v4.2.x to work around the issue, you have to overwrite those private keys again after upgrading to v4.3.0
```
fdevans: "If you added/changed any private keys in v4.2.x"

Also, is it ONLY private keys?  What about Public Keys and Passwords?

fdevans (changes requested):

docs/upgrading/upgrading-to-4.3.md
jtobard: We tested password, public and private keys, they were all affected.

docs/upgrading/upgrading-to-4.3.md
jtobard: Upgrading for any version 4.1 or earlier get the same problem

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -24,4 +24,9 @@ Steps to overwrite key storage:
 
 - Go to the key storage page
 - Select the private key need to be overwritten then click the Overwrite Key button in the toolbar
-- Follow the same steps as [Add a Private Key](/manual/system-configs.html#key-storage) to overwrite the keys
\ No newline at end of file
+- Follow the same steps as [Add a Private Key](/manual/system-configs.html#key-storage) to overwrite the keys
+
+## For Github SCM users:
```
fdevans: Is this specific to the 4.3 release or just timing on the GitHub side.  If the same issue applies to all versions of Rundeck this note needs to go on the SCM Plugin docs page, not the Upgrading notes.

ehe-pd: This Github issue applies to all versions of Rundeck.

gschueler: if there is no fix for this in 4.3, it doesn't belong here

docs/upgrading/upgrading-to-4.3.md
```diff
@@ -1,5 +1,32 @@
...
+Steps to overwrite key storage:
+
+- Go to the key storage page
+- Select the private key need to be overwritten then click the Overwrite Key button in the toolbar
+- Follow the same steps as [Add a Private Key](/manual/system-configs.html#key-storage) to overwrite the keys
+
+## For Github SCM users:
+[Github has stopped supporting insecure algorithms](https://github.blog/2021-09-01-improving-git-protocol-security-github/#when-are-these-changes-effective), and our current SSH implementation doesn't support the algorithms required by Github so you cannot use `Private Key` method to authenticate Rundeck client to Github. As a workaround, You still can use the `Password` authentication method with the Github `Personal Access Token` to connect to your Github repos.
+
+We are working on upgrading the SSH implementation to solve this issue in a future release.
```
gschueler:
```suggestion
```

docs/upgrading/upgrading-to-4.3.md
fdevans: Greg is right.  As I said above, it belongs on the feature page as a warning box.

docs/manual/projects/scm/git.md
```diff
@@ -21,6 +29,15 @@ _Project Settings > Setup SCM_
...
 
 **Git url** and **Branch** are the common repository settings.
+- For `HTTP/HTTPS` urls use the form: `http[s]://user@host.xz[:port]/path/to/repo.git`
+:::tip
+Many SCM systems provide a "clone" url for http[s] in the form: `http[s]://host.xz/path/to/repo.git`, to make use of this in the default git plugin it is necessary to add it the scm user with an `@` sign in the url: `http[s]://scm-user@host.xz/path/to/repo.git`
+:::
+
+- For `SSH` urls use the form: `ssh://user@host.xz[:port]/path/to/repo.git`
+:::tip
+Many SCM systems provide a "clone" url for ssh in the form: `git@host.xz:path/to/repo.git`, to make use of this in the default git plugin it is necessary to prepend `ssh://` and replace the `:` with a `/` in the url: `ssh://git@host.xz/path/to/repo.git`
```
ehe-pd, line 39: Can we simplify the tip to:

:::tip
Some SCM systems provide repo URLs in a custom format, please convert that format into the exact format as `ssh://user@host.xz[:port]/path/to/repo.git`. 

E.g. Github uses the format `git@github.com:path/to/repo.git` for SSH clone. You need to convert this URL into `ssh://git@github.com/path/to/repo.git`
:::

docs/manual/projects/scm/git.md
```diff
@@ -62,6 +83,15 @@ Changing this value modifies the file definition and files need to be pushed aga
...
+:::
```
ehe-pd, line 94: This section has duplicated content from lines #32-#39. Can we just reference that section?

fdevans:
Closing this as the upgrade instructions aren't needed.  The steps to remediate the issues in 4.2 will be covered elsewhere.

PR #985, opened 2022-06-04
Added setting to specify the largest request size the server can acce…

avelasquezr:
Fixed: https://github.com/rundeckpro/rundeckpro/issues/2243
**File Size Max**
_Specifies the size of the largest request the server can accept_

### DIFF
--- a/docs/administration/configuration/config-file-reference.md
+++ b/docs/administration/configuration/config-file-reference.md
@@ -614,6 +614,11 @@ File size in bytes or with a suffix of `k`,`m`,`g`,`t` (kilo,mega,giga,tera).
 ```properties
 rundeck.fileUploadService.tempfile.maxsize=200M
 ```
+Max controller file size. File size in bytes. In example: `4096000000` is equivalent to `4g`.
+```properties
+grails.controllers.upload.maxFileSize=4096000000
+grails.controllers.upload.maxRequestSize=4096000000
+```
 
 Max temp file expiration (duration in milliseconds).
 The uploaded file will be removed if not used as a job option within ths time period.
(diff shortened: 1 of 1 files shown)


avelasquezr:
This issue is a configuration issue. In the documentation you only find the change of the properties as an environment variable in Docker.

What has been added is for the section that generates the configuration files "rundeck-config.properties".
image

In addition to this, while testing was done, for the war files, the JVM values should be set, both -Xms and -Xmx

docs/administration/configuration/config-file-reference.md
```diff
@@ -624,6 +624,23 @@ because the File Upload and Job Run are performed as separate steps.)
...
 ```
+### File Size Max
+Specifies the size of the largest request the server can accept
+
+For 1GB ~ 2GB files, is recommended to set:
+```properties
+grails.controllers.upload.maxFileSize=4096000000
+grails.controllers.upload.maxRequestSize=4096000000
+rundeck.fileUploadService.tempfile.maxsize=4096000000
+rundeck.fileUploadService.tempfile.expiration=600000
```
fdevans: Is this in seconds, milliseconds?  Let's add a line somewhere in this update that clarifies what the values represent for each setting.

fdevans (changes requested):

docs/administration/configuration/config-file-reference.md
avelasquezr: **Hi**, the added module is removed and an existing one is used that deals with the same thing. Descriptions are added that represent each value in the configurations.

### Old configuration:
image

### New configuration:
image

fdevans (approved):

PR #987, opened 2022-06-06
Add another example about how to save/read a license from custom folder

Jesus-Osuna-M:
Fix:https://github.com/rundeckpro/rundeckpro/issues/1576

Problem: The documentation was not clear about how to read/save a license on a folder different than root, a change was made with the PR https://github.com/rundeckpro/rundeckpro/pull/2566 and a new property needs to be filled in order to read/save the license from the specific folder

Solution: Use an existing but not documented property to declare the path of the license. 

Propertie: rundeck.license.file


fdevans:
If this is fixed in 4.4.0 then please send the PR against the `4.4.0` branch so it gets released with that version release.

docs/administration/license.md
```diff
@@ -79,7 +79,7 @@ storage tree config options.
 
 Turn on license storage tree usage then provide your configuration using the prefix: `rundeck.license.storage`
 
-Example storing the license in Amazon S3
+Example storing the license on Amazon S3 root folder
```
fdevans, line 82: What would need to change to pull from a different folder than root?

fdevans (approved):

PR #995, opened 2022-06-09
Example value modified on porperty rundeck.gui.paginatejobs.max.per.page

Jesus-Osuna-M:
Ticket: RUN-313

Problem: property rundeck.gui.paginatejobs.max.per.page had a boolean as example value.

Fix: property rundeck.gui.paginatejobs.max.per.page is now setted as integer.


fdevans:
Please don't merge PRs until approvals are completed.

PR #1016, opened 2022-06-28
Add RD_INSECURE_SSL_NO_WARN env var suggestion

saidmasoud:
Version 2.0.0 of the `rundeck-cli` will fail with an NPE if `RD_INSECURE_SSL=true` is set. https://github.com/rundeck/rundeck-cli/issues/453 suggests also setting `RD_INSECURE_SSL_NO_WARN=true` to address the issue (which it does), so adding this to the docs so that others don't have trouble in the future.

### DIFF
--- a/docs/rd-cli/configuration.md
+++ b/docs/rd-cli/configuration.md
@@ -135,7 +135,7 @@ See [SSL Configuration](./ssl.md)
 
 To disable *all* SSL certificate checks, and hostname verifications:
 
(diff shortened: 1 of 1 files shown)


gschueler:
@saidmasoud rd-cli 2.0.1 was just released, which should fix this issue.  let us know if you still have a problem

saidmasoud:
verified, thanks!

PR #1028, opened 2022-07-14
Run 1042 - Health check running out of threads

ronaveva:
fix https://github.com/rundeckpro/rundeckpro/issues/2612

Documentation for old health check properties that changes task executor pool and queue size

### DIFF
--- a/docs/manual/healthchecks.md
+++ b/docs/manual/healthchecks.md
@@ -91,11 +91,11 @@ Each Health Check will result in a Health Status:
 
 	![Health Checks - Unauthorized Warning](~@assets/img/healthchecks-unauthorized.png)
 
-4. Once Access Control is configured, the checks should be showing up and healthy:
+5. Once Access Control is configured, the checks should be showing up and healthy:
 
 	![Health Checks - Healthy checks](~@assets/img/healthchecks-healthy-checks.png)
 
-5. Return to the "Project Settings... > Edit Nodes" page. Under "Enhancers" click "Add a new Node Enhancer"  and choose "Health Status".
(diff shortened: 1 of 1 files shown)


docs/manual/healthchecks.md
```diff
@@ -91,21 +91,31 @@ Each Health Check will result in a Health Status:
...
 
 	![Health Checks - Add Health Status Enhancer](~@assets/img/healthchecks-add-health-status-enhancer.png)			
 
-6. Visit the "Nodes" link in the Sidebar. There will be healthy status indicators for the nodes:
+7. Visit the "Nodes" link in the Sidebar. There will be healthy status indicators for the nodes:
 
 	![Health Checks - Node Health Status UI](~@assets/img/healthchecks-health-status-ui.png)			
+	
+:::tip
+If you have a very large node list, you might need to increase the thread pool size value for the healthcheck process.
```
fdevans: How would a customer know their Health Check process is running out of threads?  Is there an error they can look for?  Is there a number of nodes they should consider this for?

ronaveva: Yes, there's an error in the service log that can easily be identified by the user.

`Caused by: org.springframework.core.task.TaskRejectedException: Executor [java.util.concurrent.ThreadPoolExecutor@cbbb8a8[Running, pool size = 25, active threads = 25, queued tasks = 500, completed tasks = 8237]] did not accept task: com.google.common.util.concurrent.ListenableFutureTask@36f67998`

Related to the values i would recommend: queue+pool >= nodes

I'll add this information to the docs

fdevans (approved):

PR #1042, opened 2022-08-01
Moving cluster warning on top of upgrade page

### LINKED ISSUE #1041: Cluster upgrade guide is inside windows portion of the doc
This instruction is not exclusive to windows:
 **Before performing this upgrade, we highly recommend you stop all cluster members and do the following steps, one cluster member, at a time.**
 
 It should be on top of the upgrade guide instead of inside the Windows as a service section.

**Source page**
https://docs.rundeck.com/docs/upgrading/upgrading.html#windows-as-a-service

jtobard:
Fix #1041

### DIFF
--- a/docs/upgrading/upgrading.md
+++ b/docs/upgrading/upgrading.md
@@ -1,6 +1,10 @@
 # General Upgrade Guide
 
 
+## Cluster upgrade
+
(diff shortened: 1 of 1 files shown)


fdevans:
I did a more comprehensive update for this.

PR #1053, opened 2022-08-16
RUN-1128: add loop script plugins docs

ltamaster:
Documentation for the new loop script plugin 
https://github.com/rundeckpro/rundeckpro/pull/2625


fdevans (approved):
approved after a few minor fixes

PR #1066, opened 2022-08-23
Updated details about Global Variable Job Step with new example

Jesus-Osuna-M:
Problem: The explanation of how to use the global variable was not clear, and missed an example about how to use it.

Solution: Complete the missing information and add an example with images about how to use it properly.


fdevans:
This looks great

PR #1078, opened 2022-09-05
RSE-88: Added reference for notification context in job workflows page

L2JE:
Just added a reference to notification plugins main page and a note to remind that not all notification plugins have the same context variables.


fdevans:
@L2JE Let's chat about this and I can help write clarifications where needed.

fdevans (changes requested):
@fdevans will help re-write after chatting with @L2JE

fdevans (approved):

PR #1092, opened 2022-09-24
Progress Badge log filter enhancements

jsboak:
Overhauled the doc to include the new enhancements and walk through an example use-case.


fdevans (approved):
Approved after rebase against 4.7.0 branch so width fix isn't reverted.

PR #1097, opened 2022-09-28
Update mysql.md

liberodark:
Fix confusion

image


fdevans:
Thanks for the suggestion @liberodark but I'd prefer to leave this as a mock host name than a single word value.

PR #1108, opened 2022-10-25
RSE-211: Improve GH script step plugin documentation

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-211
Depends on: https://github.com/rundeckpro/rundeckpro/pull/2768

#### Problem
For users that never used the plugin is not easy to understand what every field actually means

#### Solution 
Changed plugin field descriptions and names, and extended the docs page of the plugin with examples.

image


alexander-variacode (approved):
LGTM

fdevans:
Is this related to the 4.9 release?  If so the PR needs to go against the 4.9.0 branch so it's merged when that is released.

fdevans (approved):

PR #1117, opened 2022-11-08
doc for changing local users passwords

jsboak:


jsboak:
This will allow us to direct users to this doc from the Solution Guide and Tours, rather than having to build a Tour just for this workflow - given that it's only applicable to local users upon their first login.

fdevans:
Probably want to clarify it's Enterprise local users created with the user wizard. Not the other local types.

fdevans (approved):

PR #1118, opened 2022-11-09
RSE-207 Fix: Missing documentation for runner API 

Jesus-Osuna-M:
Problem: There is no documentation about how to manage the Runner via API.

Solution: Add to the rundeck-api.md information about the avaiable methods to manage the Runner, how to build the request and the expected response.


fdevans:
@Jesus-Osuna-M These endpoints may change with some planned enhancements.  We are going to hold off on officially documenting or releasing these details until a later date.

PR #1126, opened 2022-11-28
Update Rundeck as a Windows service

dldrk:
Updating creation of Rundeck service on Windows using up to date standard Apache utilities, as NSSM binary's latest release was on 2014-08-31.


MegaDrive68k (approved):
Tested on Windows 10, it works. LGTM.

PR #1131, opened 2022-12-02
Updated instructions on password authentication with JSESSIONID

bproverb:


fdevans (approved):

fdevans (changes requested):
If this is permanent we should note it in the Release Notes and probably an "Upgrading" note.

fdevans (approved):

PR #1135, opened 2022-12-06
Update advanced-run-job.md

lcanavosio:
added rundeck.feature.eventStore.enabled property in FAQs


fdevans (changes requested):
we need blank lines between 259 and 260 and between the section header (currently 260) and the content (261)

fdevans:
We should also note if this can be set using Configuration Management using a custom setting for ENT customers.

lcanavosio (commented):
for review

fdevans (approved):

PR #1136, opened 2022-12-06
Update config-file-reference.md

mauriciogc-sudo:
Adding a NOTE on the "Job file Option Uploads" related to a shared folder.

### DIFF
--- a/docs/administration/configuration/config-file-reference.md
+++ b/docs/administration/configuration/config-file-reference.md
@@ -647,6 +647,9 @@ java -Xms4g -Xmx8g -jar rundeck.war
 It is also possible to reconfigure the default path for files according to [this page](/developer/14-file-upload-plugins.html#about) using the property in this form:
 
 `framework.plugin.FileUpload.filesystem-temp.basePath=/desired/path`
+
+'''NOTE : If you are using the remote execution policy, you might to know that this “Upload File” option will work properly if you have on your rundeck cluster an NFS, or other shared folder. S3 is not supported yet.
+
 :::
 
 ### Job YAML format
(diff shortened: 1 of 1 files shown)


fdevans:
@mauriciogc-sudo This needs a bit more clarity and some edits for how it's worded.  Can you provide more details what the goal of this is?

mauriciogc-sudo:
@fdevans Hi, sure. The context is this:
If you have a rundeck cluster, remote execution policy, lb , many jobs with the Upload option file, and those jobs were created on only one cluster member, every time that someone login to rundeck, the lb will connect you to every cluster member, so when you try to upload a file and you are not on the owner member, the upload will fail with an error like "rundeck can not find the file". So, this can be fixed using an nfs on every cluster member, with this shared folder the upload option file will work no matter the lb connected to.
I talked with dev and told me that this is not supported, but works. The idea was to add a note for this if you have a cluster with this configuration.
Makes sense?
We have the customer CGS with this problem. They asked me why does not say anything about this on the docs.

fdevans:
@wayfaringson Can you collaborate with @mauriciogc-sudo  on this?

wayfaringson:
Will do.  Chatted with him earlier and we'll come up with clearer language for this situation.

wayfaringson (changes requested):
How about this @mauriciogc-sudo 

NOTE:
In a clustered mode, it is important to note that this option by default will only upload the file to a single cluster member (whichever server you are uploading the file on).  To address this in a better way, it is recommended that you set the desired path to a directory that is shared across all cluster members.  This could be and NFS share or something like it.  By doing so, when a file is uploaded, it will be available for all cluster members.

mauriciogc-sudo:
@wayfaringson yeah, sounds good, sound better.

Thanks

fdevans:
@mauriciogc-sudo Are you going to update the PR?

mauriciogc-sudo (commented):
We changed the Note related to NFS.

fdevans (approved):

fdevans:
Approving and merging, then will fixup typos since I don't have access to the fork.

wayfaringson (approved):
Looks good.

PR #1138, opened 2022-12-09
fixing url for sample k8s job

jsboak:
Fixing URL so that asset can be pulled from GH correctly


fdevans (approved):

fdevans:
This will fix it for now, but this isn't sustainable as we do change the main branch when major versions change.  We need to figure out how to store these in the actual docs site content.

jsboak:
agreed - let's discuss this soon

PR #1145, opened 2022-12-14
RSE-241: Using wrong charset to print output on windows 12 japanese nodes

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-241
Depends On:  https://github.com/rundeck-plugins/py-winrm-plugin/pull/91

Mention new node atribute `output-charset` that can be used from any executor to render outputs. 

This will be initially used by the pywinrm node executor since the depending PR.


fdevans:
Reminder for Forrest: Before merging change dest branch to 4.10.0 which will be available after 4.9.0 release.

fdevans (approved):

PR #1150, opened 2022-12-22
add grant priviles on public schema

### LINKED ISSUE #1149: missing privileges on public schema
**Describe the bug**
missing privileges on public schema, where you can receive the following error after start rundeck or modify/create any record on database:

```log
SQL Error [42501]: ERROR: permission denied for schema publicPosition: 14
ERROR: permission denied for schema publicPosition: 14
ERROR: permission denied for schema publicPosition: 14
```

**Source page**
https://docs.rundeck.com/docs/administration/configuration/database/postgres.html

jgarces-pd:
add new schema privileges for pgsql 15+ (can be applied on lower versions too)
Fixes: https://github.com/rundeck/docs/issues/1149

### DIFF
--- a/docs/administration/configuration/database/postgres.md
+++ b/docs/administration/configuration/database/postgres.md
@@ -14,17 +14,25 @@ If it is not running, start Postgres (with `service postgresql-<version> start` 
 
 Switch to 'postgres' user and use the 'psql' commandline tool to access the db:
 
-    $ su postgres
-    $ psql
+```bash
+$: su postgres
+$: psql
+```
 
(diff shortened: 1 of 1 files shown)


fdevans:
We will likely get the question at some point, Does it really require "ALL" privileges?

jgarces-pd:
> We will likely get the question at some point, Does it really require "ALL" privileges?

not sure, but I'm following the docs:

postgres=# create user rundeckuser with password 'rundeckpassword';
postgres=# grant ALL privileges on database rundeck to rundeckuser;

We already have that one. Not sure if we should support custom privileges.

@jtobard

fdevans (approved):

PR #1156, opened 2023-01-10
Enhanced Plugin Configuration (Beta)

jsboak:
This outlines the basic configuration methods for plugins at the project and system level.  Geared towards first-time/basic users, this doc describes how to set properties for plugin suites at the project/system levels using the recently enhanced UX for this operation (leveraging the beta of the PluginGroup feature).


fdevans:
Are there other areas we need to change to direct them here for this configuration?  I have a feeling there are docs in quite a few places that use the "old way".  

At a minimum let's highlight the new steps on all the PagerDuty plugin setups.

https://docs.rundeck.com/docs/manual/workflow-steps/pagerduty.html#pagerduty-workflow-steps-enterprise for example

jsboak:
@fdevans i've updated PD plugins page. Other spots throughout docs should be part of the general enhancements to the new-user-onboarding that Nathan's taking on.

fdevans (approved):
Approved as a start, but i think we have a lot more we need to do to point between various plugin references.

PR #1160, opened 2023-01-13
Update Yum repo info

gschueler:
add new signing key
set gpgcheck=1

review instructions
----

you can use redhat/ubi8 image
```
docker run -it redhat/ubi8
```

use the "manual instructions"


fdevans (approved):

ehe-pd (approved):
Tested. Installation works as expected.

PR #1163, opened 2023-01-24
RSE-255 Fix: Problems with LDAP integration with a huge number of groups

Jesus-Osuna-M:
Problem: So far Rundeck does not support pagination on Active Directory-LDAP authentication and the maximum amount of results that Rundeck is able to get from AD is specified on the Active Directory side on MaxPageSize. 

Solution: Add a tip on documentation to prevent clients from exceeding the maximum results specified on their Active Directory 

More info: https://pagerduty.atlassian.net/browse/RSE-255

image

### DIFF
--- a/docs/administration/security/authentication.md
+++ b/docs/administration/security/authentication.md
@@ -429,6 +429,10 @@ The [PAM](#pam) section is a useful comparison as it uses the same method to com
 
 ### Active Directory
 
+:::tip
+The maximum quantity of results that Rundeck is able to retrieve from Active Directory is limited to the value specified for the MaxPageSize parameter within the Active Directory configuration.
+:::
+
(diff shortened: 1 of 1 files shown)


docs/administration/security/authentication.md
```diff
@@ -429,6 +429,10 @@ The [PAM](#pam) section is a useful comparison as it uses the same method to com
 
 ### Active Directory
 
+:::tip
+The maximum quantity of results that Rundeck is able to retrieve from Active Directory is limited to the value specified for the MaxPageSize parameter within the Active Directory configuration.
```
fdevans, line 433: This needs more details. Where is this setting inside Active Directory?  How would they change it?  Preferably point to Microsoft Docs that discuss this setting.

fdevans (changes requested):

PR #1165, opened 2023-01-25
RSE 206: Thycotic Storage Plugin settings do not work when they are set in the GUI ( System Configuration)

DarwisNarvaezDev:
# RSE 206: Thycotic Storage Plugin settings do not work when they are set in the GUI ( System Configuration)
An issue was reported stating that the Thycotic plugin's settings dont work if the properties were set through Configuration Manager, later we found that maybe it's a matter of documentation.

## The Problem
Previously, the docs say that the "recommended" way to set the props via Configuration Manager was to set the Thycotic Plugin as a **main** storage provider; that results in a bad usage of the plugin, because we are squashing the rundeck-config.properties file configuration, also, when we restart rundeck, Thycotic's secrets could not be seen anymore, because rundeck takes the file's configuration.

## The Solution
Update the info, warn about the password exposure, and explain a little bit more about the ```removePathPrefix=true``` property


docs/manual/key-storage/storage-plugins/thycotic-storage.md
```diff
@@ -12,21 +12,28 @@ Process Automation users have access to the Thycotic Storage Plugin which can be
 
 Below is an example configuration, which can be configured using the *System Configuration* module. Add each setting as a configuration entry.
 
-![Thycotic Configuration](@assets/img/thycotic-config-screen.png)
+:::warning
+Doing it via Configuration Manager Plugin (GUI way to write configuration) will expose sensible fields such as the password. We recommend to use "$RDECK_BASE/rundeck-config.properties" file.
```
fdevans: This is not the recommendation.  We can enter a ticket to protect the Thycotic password, but setting via rundeck-config is still not the suggested method.

docs/manual/key-storage/storage-plugins/thycotic-storage.md
```diff
@@ -12,21 +12,28 @@ Process Automation users have access to the Thycotic Storage Plugin which can be
...
+:::
+
+![Thycotic Configuration](@assets/img/thycoticConfiguration.png)
+
+**Note:** if user use Thycotic Storage Plugin as a second storage provider, and its path preffix is set to ```true```, the keys will be downloaded to a separate folder within "keys" folder, exhibit in the following:
```
fdevans: This doesn't need a special "Note".  It's a common setting across all the Key Storage plugins.

docs/manual/key-storage/storage-plugins/thycotic-storage.md
```diff
@@ -12,21 +12,28 @@ Process Automation users have access to the Thycotic Storage Plugin which can be
...
 ```
-rundeck.storage.provider.1.type=thycotic-storage
-rundeck.storage.provider.1.path=keys
-rundeck.storage.provider.1.removePathPrefix=true
-rundeck.storage.provider.1.config.username=thycotic-username
-rundeck.storage.provider.1.config.password=password
-rundeck.storage.provider.1.config.address=https://example.secretservercloud.com
-rundeck.storage.provider.1.config.allowSelfSignedCert=true
-rundeck.storage.provider.1.config.maximumResources=100
+rundeck.storage.provider.2.type=thycotic-storage
```
fdevans: Why did all these change to `.2`?

fdevans (changes requested):

docs/manual/key-storage/storage-plugins/thycotic-storage.md
DarwisNarvaezDev: ok! i'll remove it

docs/manual/key-storage/storage-plugins/thycotic-storage.md
DarwisNarvaezDev: ok, i'll remove it

docs/manual/key-storage/storage-plugins/thycotic-storage.md
DarwisNarvaezDev: Please take a look at this: https://pagerduty.atlassian.net/browse/RSE-206 ; i think that issue came because the users will not use the plugin properly if the steps in this docs are strictly followed.

fdevans: That's where that index setting can cause issues.  It is brittle, in that they have to start at 1 and increase properly.  We should put in [index] and have the customer choose the right value for their situation.  If there isn't a good explanation of this setting in the docs let's find a spot for it.

DarwisNarvaezDev: Sounds good, i can give it a try.

fdevans (approved):

PR #1171, opened 2023-02-09
Add Global Log Filters content to Docs

fdevans:
Opening initial update for @jtobard to review and add additional details.

Open Questions/Tasks:

- Does this work with Configuration Management?
- What LogFilters work for the Project Global or Rundeck Global settings?  If all of them are possible we need to know the proper setting values for each type.  Can be documented on the type page or noted that they don't work with Global.
- We should file a follow-up ticket to change the "Global Log Filter" words on the Job Edit page to "Job Log Filter - All Steps"


jsboak (approved):
lgtm!

PR #1184, opened 2023-02-21
RUN-1412: Runner Management API docs

ehe-pd:
Add Runner Management API documentations. The document is converted from the system-generated OpenAPI Yaml file.


fdevans:
@ehe-pd Does the API Version change with this release?  If so there are likely a few other places we need to update.

ehe-pd:
> Does the API Version change with this release?

There is no version change. Those APIs are all v41 and v42 from Runner Management

gschueler (approved):

fdevans (approved):
Approving these, but we are holding off on merging until we finalize plan for other Runner docs updates.

PR #1190, opened 2023-03-07
ServiceNow plugin suite

jsboak:
docs for https://github.com/rundeckpro/rundeckpro/pull/2882


docs/manual/plugins/servicenow-plugins-overview.md
```diff
@@ -0,0 +1,83 @@
...
+
+Use the following steps to configure authentication for the ServiceNow plugins for a specific project.
+
+1. In the specific project, click on **Project Settings** in the lower left.
+2. Click on **Edit Configuration** then click on **Plugins**.
+   ![Plugin Suite Project Settings](@assets/img/plugin-groups-project-settings.png)<br>
+3. Click on **+PluginGroup**.
+4. Select **ServiceNow** from the list.
+5. Add your ServiceNow instance URL to the **Server URL** field.
+6. Place an active ServiceNow username into the **User name** field.
```
fdevans: user with sufficient rights to perform the actions the automation will be doing.

fdevans (changes requested):
I don't see any changes/updates to the existing ServiceNow plugin docs related to this.  I think we should be reviewing those and pointing customers to configure using these steps.  Also need to add any info needed about "adding a Note" that was added in 4.11

jsboak:
@fdevans were any docs created for these new plugins? i didn't see any

fdevans:
I don't think there was docs done yet for the "Notes" feature

fdevans (approved):

PR #1196, opened 2023-03-16
How To: RD-Terraform on EKS and Vanilla AWS

wayfaringson:
Two new Community How2 articles focused on deploying RD with Terraform on EKS and vanilla AWS.  Also includes updates to the glossary of terms to include terms for the upcoming Jobs 101 content.


docs/learning/howto/how2-terra-rd-aws.md
```diff
@@ -0,0 +1,299 @@
+# Deploying Rundeck on AWS with Terraform
```
fdevans, line 1: I think we should note that this works (or doesn't) on both OSS and Enterprise versions somewhere just for clarity.  In other docs we call our the two versions and highlight differences.

docs/learning/howto/how2-terra-rd-eks.md
```diff
@@ -0,0 +1,176 @@
+# Deploying Rundeck on AWS EKS with Terraform
```
fdevans, line 1: Same as other comment.  Identify whether this is Enterprise Only or available for OSS

fdevans (changes requested):

docs/learning/howto/how2-terra-rd-aws.md
wayfaringson: Just added updated language in both overviews that specifically include Process Automation as covered by these instructions

fdevans (approved):

PR #1197, opened 2023-03-21
RSE-316: SCM autopull and repo usage

ronaveva:
- It specifies that the "auto pull" does not import changes into rundeck.
- Adds a tip to try to avoid usage of same repo and branch for export and import on a single project. This due to export and import getting each other out of sync by having different last commit id.

This doc does not require code changes


fdevans:
If this is for 4.12 please make PR against `4.12.0` branch.

fdevans (approved):

PR #1206, opened 2023-03-30
RSE-345: Add Secure Your Deployment section to runner installation page

ehe-pd:
image


ltamaster (commented):
I think we should add what was described in the ticket about setting the custom path using the configurations, like: 

```
runner:
    rundeck:
        overrideTempDir: true
    dirs:
        tmp: C:\path\to\tmp-dir-w-custom-security
```

or using ` file-copy-destination-dir=C:\\path\\Temp` in the Local Node

ltamaster:
also the configurations can be set as a java env parameters, eg:
```
java -jar -Drunner.rundeck.overrideTempDir=true -Drunner.dirs.tmp=C:\path\to\tmp-dir-w-custom-security pd-runner.jar
```

jsboak (approved):
looks good. we should have a separate PR for adding JVM customization, such as custom tmp directory and custom logging

PR #1209, opened 2023-04-07
Enhanced Remote URL Job Options

jsboak:
Documentation for additions to the Remote URL Job Options:
1. Auth methods: Basic, API Key, and Bearer Token
2. Accessing and leveraging nested JSON lists for the option values


docs/manual/job-options.md
```diff
@@ -446,31 +446,33 @@ The path must indicate a stored `password` entry in the storage facility.
 
 ![Storage Path for Secure Option](~@assets/img/jobs-options-secure-storage-path.png)
 
-## Remote option values
+## Option Model Provider
```
fdevans, line 449: Can we find a more descriptive way to describe this?  Happy to help brainstorm.

docs/manual/job-options.md
```diff
@@ -446,31 +446,33 @@ The path must indicate a stored `password` entry in the storage facility.
...
+This is useful in a couple of scenarios when Process Automation is used to
+coordinate processes that depend on other systems:
+
+- Deploying packages or artifacts produced by a build or CI server, such as Jenkins:
   - A list of recent Jenkins build artifacts can be imported as Options data, so that a User can pick an appropriate package name to deploy from a list.
 - Selecting from a set of available environments, defined in a CMDB
 - Any situation in which input variables for your Jobs must be selected from some set of values produced by a different system.
 
-## Option model provider
+When defining the Job in YAML, XML, or JSON the `valuesUrl` is specified for an Option, then the model of
```
fdevans, line 467: I think the Remote URL only supports JSON so let's leave the YAML, XML parts out

docs/manual/job-options.md
```diff
@@ -446,31 +446,33 @@ The path must indicate a stored `password` entry in the storage facility.
...
-Option model providers are configured on a per-Option basis (where a Job may have zero or more Options).
+    <option valuesUrl="http://site.example.com/values.json" ... />
 
 ### Requirements
 
-1. Options model data must be [JSON formatted](http://www.json.org).
-2. It must be accessible via HTTP(S) or on the local disk for the Rundeck server.
+1. The returned data must be [JSON formatted](http://www.json.org).
+2. It must be accessible via HTTP(S) or on the local disk for the Process Automation or Rundeck server.
 3. It must be in one of two JSON structures, _either_:
```
fdevans: "It must be in or contain a..." Since we can now support something below the root doc

docs/manual/job-options.md
```diff
@@ -543,7 +549,25 @@ Name Value List with default selections:
 ]
 ```
 
-### URL connection parameters
+#### Using Nested JSON Elements
```
fdevans, line 552: Let's include a link to an official (or popular) JSON Path resource for deeper reference.

fdevans (changes requested):

docs/manual/job-options.md
jsboak: yes - i'd like to change it too, but this is now the hyperlink that is used in the product to link to the docs. would need separate PR to change this and that

docs/manual/job-options.md
jsboak: this is referring to the Job Definition, not the returned data

docs/manual/job-options.md
fdevans: We have ways to do redirects within the app.  I'll have to see if it will work with an anchor though.

fdevans (approved):

PR #1231, opened 2023-05-10
Fix architecture link

wayfaringson:
Re-pushing updated cluster landing page and architecture PDF with working links in it.

Currently, the site has an old version of the page and the PDF has no working links in it at all, due to a problem with the PDF conversion process.


jsboak (approved):
LGTM!

PR #1237, opened 2023-05-10
Update README.md

wayfaringson:
Fixed url so it actually works now


jsboak (approved):
that _should_ work ;-)

PR #1245, opened 2023-05-24
New property for db migrations

ronaveva:
New property for db migrations when using mysql DB
Needs merging of this PR https://github.com/rundeck/rundeck/pull/8215


fdevans (changes requested):
I believe this is changing per a thread from today.

PR #1248, opened 2023-05-25
Page Rewrites

fdevans:


docs/administration/install/docker.md
```diff
@@ -1,5 +1,7 @@
 # Running Docker Images
 
+Running Rundeck in Docker offers several advantages, including portability across different environments, consistency in application behavior, isolation from other applications, scalability for handling increased workloads, and simplified deployment and management processes. Docker's containerization enables the packaging of Rundeck and its dependencies into a single container, allowing for easy distribution, deployment, and scaling. This approach ensures that Rundeck operates consistently, reduces compatibility issues, and provides flexibility for customers to run Rundeck in their preferred environments.
+
```
jsboak, line 4: do we really want all this extra text? seems like information we'd put into a white-paper rather than into the tech-docs.

fdevans: It's just color text that I was using to provide more content on the page.

jsboak (approved):
LGTM

PR #1250, opened 2023-05-29
RSE-405: New documentation is proposed to clarify automatic pull.

avelasquezr:
Before change: Ref: https://docs.rundeck.com/docs/manual/projects/scm/git.html#git-repository-configuration-2
image
And:
image

After change:
image
And:
image


fdevans:
When a Doc PR is tied to a specific release please make the PR against the branch with that version number so it gets merged when the release happens.

fdevans (approved):

PR #1264, opened 2023-06-29
Defined required permissions for token

simon-c-msc:
Added the minimum required permissions for the token to get the plugin to work with Rundeck

### DIFF
--- a/docs/manual/integrations/servicenow-app.md
+++ b/docs/manual/integrations/servicenow-app.md
@@ -27,7 +27,34 @@ After the app has been installed in your ServiceNow&reg; instance there will be 
 
 > Note: The ServiceNow login account will need the `admin` role and `x_runde_rundeck_it.rundeck_app_user` roles to configure these settings.
 
-1. Create an [API Key](/manual/10-user.md#user-api-tokens) in Rundeck.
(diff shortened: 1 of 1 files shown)


docs/manual/integrations/servicenow-app.md
```diff
@@ -27,7 +27,34 @@ After the app has been installed in your ServiceNow&reg; instance there will be
 
 > Note: The ServiceNow login account will need the `admin` role and `x_runde_rundeck_it.rundeck_app_user` roles to configure these settings.
 
-1. Create an [API Key](/manual/10-user.md#user-api-tokens) in Rundeck.
+1. Create either a [User API Key](/manual/10-user.md#user-api-tokens) or a [Static API Token](/administration/configuration/config-file-reference.md#static-authentication-tokens-for-api-access)
+    1. The minimum required permissions for the token are:
+        1. Read all or specific projects
+        2. Read all or specific webhooks in the projects necessary
+        3. Post for all or specific webhooks
+    4. Example ACL:
```
fdevans, line 35: Need to specify that this ACL would allow access if they create and add users to a group called `snow_webhooks`.

(Note: we should not use "snow" but "sn"

fdevans (changes requested):

fdevans:
merging to a local branch so we can make improvements.  Thanks for the PR @simon-c-msc

PR #1266, opened 2023-07-09
Add User Class Mgr Endpoints to API docs

cwaltherf:
adding user class endpoints


fdevans:
changed base to a local branch and merging to make edits and get reviews prior to publishing.

PR #1269, opened 2023-07-11
Updated SSH key generation command

tomerarazy:
The current command: `ssh-keygen -t key-type -m PEM -f key-pair-name -C ""` is no longer working (at least with GitLab). 
As suggested in [here](https://github.com/rundeck/rundeck/issues/7923#issuecomment-1254460493) `ed25519` is the way to go.


fdevans (approved):

fdevans:
Thank you @tomerarazy !!

PR #1278, opened 2023-07-17
Update authentication.md - Important note for --testauth flag usage w…

dldrk:
Important note for --testauth flag usage with passwords


fdevans:
I'd like to understand more background on this before we merge this as an update.

dldrk:
It's a heads up by experience after testing the --testauth flag with bind password at rundeck-config file and not jaas file. The auth test will always fail because that flag needs the password explicitly at jaas file.

fdevans (approved):

PR #1283, opened 2023-07-27
Update resource YAML documentation to rename ssh-keypath attribute to ssh-key-storage-path

programster:
Updates the YAML node resource documentation to change the `ssh-keypath` attribute name to `ssh-key-storage-path`. 

When testing, I found that `ssh-keypath` did not work. However I found `ssh-key-storage-path` documented in the[ JSON documentation examples](https://docs.rundeck.com/docs/manual/document-format-reference/resource-json-v10.html#examples). When I tried this different name in my node YAML definition, it worked.


MegaDrive68k (approved):
Same in my environment. Good catch @programster. Thanks.

PR #1291, opened 2023-08-02
Update resource-json-v10.md

freeterman:


fdevans:
@freeterman Can you help us understand why you think this change is necessary?  Using a port on the HostName entry isn't common for example purposes.

fdevans:
Closing until further feedback is received.

PR #1298, opened 2023-08-14
RSE-694 misleading docs on plugin blocklist

Jesus-Osuna-M:
Problem: Documentation about how to use the blocklist plugin was confusing.

Solution: Added two examples of how to use the blocklist plugin by filenameEntries, and another using the providerNameEntries, both with functional yaml examples and images.


docs/administration/security/blocklist.md
```diff
@@ -1,27 +1,57 @@
...
+    - rundeckpro-plugin-jar-name
+    - rundeckpro-plugin-jar-name
+    - rundeckpro-plugin-jar-name
 providerNameEntries:
    ProviderService:
       - provider-name
       - provider-name
+```
+:::tip
+If you intend to use the fileNameEntries or providerNameEntries, both of these properties must be declared otherwise Rundeck will fail to start.
```
fdevans: Are you trying to say that "even if one section is blank, it must still be declared" ?

Jesus-Osuna-M: Yes, that is what I was trying to explain with this paragraph, I will fix it with your suggestion.

fdevans (approved):

PR #1325, opened 2023-09-21
Fix typo 'sickness'

brmdias:
Change 'sickness' to 'stickiness'.


brmdias:
Also added an example to the Dynamic runner selection.
To avoid users getting confused with something different, the screenshots helps the users to correctly understand how to configure the job option and runner filter.

fdevans (approved):

PR #1332, opened 2023-09-28
Move `assets` and update image paths

fdevans:
@jsboak and @wayfaringson making this change to set us up for the VuePress 2 upgrade.

It moves the Assets folder and there is now a new "path" to call out to images, etc.  Please review / test that your items are working and approve.


jsboak (approved):
Ran locally and appears to operate normally

wayfaringson (approved):
Spot checks look good.

PR #1338, opened 2023-09-29
RSE-812: Doc: not importing logs from s3logstore

L2JE:
Ticket: https://pagerduty.atlassian.net/browse/RSE-740
Depends on: https://github.com/rundeck/rundeck/pull/8583

Documentation for the usage of `outputfilepath`, `isRemoteFilePath` and `execIdForLogStore` execution context variables for retrieving log files for imported executions in Log Storage Plugins.

### DIFF
--- a/docs/developer/06-logging-plugins.md
+++ b/docs/developer/06-logging-plugins.md
@@ -135,6 +135,44 @@ Note that the Map keys will not start with `job.`, simply use the variable name,
 
 In addition, for ExecutionFileStorage plugins, another map entry named `filetype` will be specified, which indicates which type of file is being stored. You need to use this filetype as part of the identifier for storing or retrieving the file.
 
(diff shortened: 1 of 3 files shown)


fdevans:
@wayfaringson - Can you have our tech-writer check this out before we merge it and touch up the wording?

fdevans:
@L2JE - Since this was opened the Docs Site has been updated.  Can you please merge the latest 4.0.x branch into this branch?  It may happen as part of the merge commit, but we need to update the images to use the new path/location.

`~@assets` goes to `/assets` and the images are placed in the `./docs/.vuepress/public/assets/img` folder.

Also, I'm not sure where the `{{{javaDocBase}}}` was previously coming from.  If we need that variable to be replaced with something let's talk.

PR #1346, opened 2023-10-08
Update linux-deb.md

x86txt:
Without the daemon-reload, systemctl reports the service cannot be found.

>➜  matt ~ sudo service rundeckd start
Failed to start rundeckd.service: Unit rundeckd.service not found.


MegaDrive68k (approved):
Right, useful on modern Debian-based distros.

PR #1354, opened 2023-10-12
Update how2-terra-rd-eks.md

jplassnibatt:
This is not related to terraform.


MegaDrive68k (approved):
Right, thanks @jplassnibatt

PR #1359, opened 2023-10-18
Fixing broken link on k8s-logs-events.md

t1agob:
The suggested link to download the job definition yaml file is broken. Looks like the file was moved to a new location.


fdevans:
This is not the proper link to this.  If placing files in the `docs/.vuepress/assets/` directory the link text should read as follows:

`[text about link](/assets/text/sample_k8s_job.yaml)`

The Vuepress will handle the links properly from there.

fdevans (changes requested):
See previous comment

t1agob:
Thanks for the feedback @fdevans. I wasn't aware the vuepress conventions. Updated now.

fdevans (approved):

PR #1370, opened 2023-11-07
Add ACL Recipes section to HowTo

fdevans:
Initial push of the ACL Recipes section.

### DIFF
--- a/docs/learning/howto/acls/group-apikey.md
+++ b/docs/learning/howto/acls/group-apikey.md
@@ -1,0 +1,32 @@
+---
+title: "Group/API Key Create User API Key"
+order: 600
+---
+
+# Group/API Key Create User API Ke
+
+## Use Case Description
+
+Assign members of a specific Rundeck Group access to create API keys associated with their username.
+
+## Code Description
+Find and replace these values with your own.
+- Group: `grp-sandbox-exec`
+
+Steps to implement are covered in the [overview page](index.md).
+
+## ACL Code
+
+``` yaml
+description: Provide access to create user API tokens
+context:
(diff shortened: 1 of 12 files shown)


docs/learning/howto/acls/group-apikey.md
```diff
@@ -0,0 +1,32 @@
+---
+title: "Group/API Key Create User API Key"
+order: 600
+---
+
+# Group/API Key Create User API Ke
+
+## Use Case Description
```
fdevans, line 8: @wayfaringson - I think we should provide more background here.  Are users not allowed to create API keys by default?

docs/learning/howto/acls/group-jobname.md
```diff
@@ -0,0 +1,50 @@
+---
+title: "Group/Project/JobName Execute Only"
+order: 500
+---
+
+# Group/Project/JobName Execute Only
```
fdevans, line 6: We should also allow read access to the project Key Storage, or note that if the jobs uses Key Storage entries it will need to use something to provide those rights.

docs/learning/howto/acls/group-apikey.md
wayfaringson: Because it is a system task, i don't believe they can do so without this (unless they are admin).

fdevans: this is a good add then

wayfaringson (approved):
I tested all of the ACL snippets and caught two things that needed correcting which are now reflected here.  As far as I am concerned, this could be merged at this point.

PR #1371, opened 2023-11-07
Update powershell.md

brmdias:
Add possibility to pass user and password from the Job to the node executor, using job options. This is documented in the UI but not in our Docs.


brmdias:
Added a tip to rename rd.bat file on Windows install.

fdevans (approved):

PR #1384, opened 2023-11-17
Adding a runner status information. 

rmeloamaro:
**Adding a runner status information as stated on the internal confluence page**
Note: Please review content with the dev team before making it publicly available as it may contain some values that were not obtained directly from the source code.


jsboak (approved):
Made clarifying changes - now good to go

PR #1391, opened 2023-11-30
Update runner-advancedsetup.md

jplassnibatt:
Adding runner + paop with self signed.


jplassnibatt:
BTW, I already tested all the examples.

fdevans:
We typically avoid documenting self-signed cert setups outside of lab builds since it's not considered "secure" by the industry as a whole.  Let's talk about a way we can document this differently to show how it's done without making it look like an official setup doc.

fdevans (changes requested):
see comment on main thread

fdevans:
Closing after discussion with @jplassnibatt.  He will re-implement as a HowTo article

PR #1409, opened 2023-12-22
Update users-overview.md

jesscrees:
Fixes typo


fdevans (approved):
Good catch @jesscrees

PR #1460, opened 2024-03-22
Update linux-deb.md

FriendlyTroll:
Updated info about adding the repo signing key as apt-key is deprecated


MegaDrive68k (approved):
That's correct. Also, I tested it and it worked in my environment.

fdevans (approved):

PR #1471, opened 2024-04-17
Update use-terraform-provider.md

brmdias:
Updating rundeck version to 0.4.5


docs/learning/howto/use-terraform-provider.md
```diff
@@ -66,7 +66,7 @@ terraform {
   required_providers {
     rundeck = {
       source  = "rundeck/rundeck"
-      version = "0.4.2"
+      version = "0.4.5"
```
fdevans, line 69: 0.4.7 is the current version

fdevans (changes requested):

PR #1472, opened 2024-04-26
Vuepress Update Packages

fdevans:
Updates VuePress packages to current versions to improve performance and stay current.


jsboak (approved):
looks good. 
needed to upgrade node to v18.19.0

PR #1508, opened 2024-07-30
Update ai-generated-runbooks.md

h4dz:
Adding information about availability in the US service region per PM Jacky Leybman


fdevans:
Thank you Hadley, but RBA SaaS only exists in the US Region today so all RBA instances should have access to the feature.  I'm not sure we want to imply that RBA SaaS might be available elsewhere.  We can either re-write it to make that clear or we'd want to leave out the implication that there are regional options for RBA.

PR #1520, opened 2024-08-20
Update user-classes.md

urbinet:


fdevans:
This isn't quite accurate as both products are named "Runbook Automation" now.  If you change the:

"This is not available in Runbook Automation."

to 

"This is not available in the Cloud offering."

I can approve it.

fdevans:
update made in #1521

PR #1531, opened 2024-09-04
RUN-2376 api/job/id/executions to return referenced executions

carlosrfranco:
The api endpoint api/jjob/{id}/executions to return the same result as activity tab at job page, including referenced executions if passing `includeJobRef=true` parameter


ahormazabal (approved):
TMLG

PR #1534, opened 2024-09-04
RUN-2825: Added info about paging limit field on ServiceNow model node source

alexander-variacode:
## Ticket related to
https://pagerduty.atlassian.net/browse/RUN-2571

## Description
To reduce the query time for nodes per page on the ServiceNow API and avoid timeouts, you can use the 'Paging limit' field to limit the number of results returned per request.


fdevans (changes requested):
Can we provide some guidance or example? Is there a default if this is not set?  What would increasing the value do to help the performance?

fdevans (approved):

PR #1562, opened 2024-10-14
Aws azure node steps

jsboak:
we bundle these aws and azure node step plugins but haven't had public docs for them yet


fdevans (approved):

fdevans (commented):
For consistency we might want to include the `.md` on the side-bar menus.  I know some use it and some don't, but i've had issues with that recently.

PR #1572, opened 2024-10-23
Update node-dispatch.md

nvd01:
corrected the below paragraph where the property "rundeck.localExecutor.disabled=true" needs to be set as JVM system property NOT in rundeck-config.properties file

"To still use the Local Node Source, but prevent the execution of commands and scripts on the Runbook Automation cluster members, set the property rundeck.localExecutor.disabled=true in the rundeck-config.properties file or DISABLED_LOCAL_EXECUTOR=true for Docker installations."


fdevans (approved):

fdevans:
Thank you @nvd01 .  We are looking into whether this is an oversight and can be improved, but for now you're correct.  It is only able to be set by system property.

PR #1573, opened 2024-10-23
Docker Solution

fdevans:
@jsboak - Along with this we'll want to update the project file.


jsboak (approved):
made a couple of minor adjustments. otherwise looks good

PR #1584, opened 2024-11-14
RPL-77: Add documentation for Kubernetes Clusters Node Step plugins

hiawvp:
Add manual entries for new Kubernetes Clusters Node Step plugins


ronaveva:
Hello @hiawvp i would prefer to remove the reference to the kubernetes plugins overview in the "Authentication" block, since it references the old python dependent plugins

Instead it could say that it uses the same authentication used to get the k8s cluster node 🤔

PR #1591, opened 2024-11-20
Update user-guide.ts - adding GCP-GKE

EcuetoPD:


fdevans:
I don't think this is in the right place.

EcuetoPD:
You're right, I already did it in the right place

PR #1603, opened 2024-12-12
Adds a Version Selector to Navbar

fdevans:
We might want to get some styling improvements from @smartinellibenedetti but this implements a drop down in the Navbar for previous versions with "Cloud" going to the latest docs.


jsboak:
This is great.  Only ask is whether the dropdown can be "backported" to the prior versions so that users can easily navigate from one version to another.  The way it's implemented now will only make the dropdown available from version 5.8.0 onwards

fdevans:
No.  There isn't an easy way to backport this as it relies on the latest vuepress template theme

jsboak (approved):
great addition!

PR #1615, opened 2025-01-31
Update index.md

rmeloamaro:
Updating placeholder {{ $apiVersion }} on Webhook entries


fdevans (approved):

fdevans:
I will approve it, but this may cause issues for customers on older versions as it will replace the value with the latest API version.  Some of these endpoints would be supported farther back.

PR #1635, opened 2025-03-06
Update ssh.md

rmeloamaro:
- Updating requirements for SSHJ as per RPL-40
- Adding info note on how to disable Bracketed Paste Mode


fdevans (approved):

fdevans:
approved, but changed base so I can make other updates to that page and content.  

We try to avoid using "you" as it's not meant to be dialog, but rather documentation.  Seems this page needs some fixes there.

PR #1647, opened 2025-04-09
Update vault-integration.md

rmeloamaro:
Adding tip for Hashicorp vault LDAP authentication


fdevans (changes requested):
Let's not document it as a bespoke tip. The settings should be spelled out more clearly.  Where does the "Userpass Mount name" setting show up?  It's not in the list above.

PR #1649, opened 2025-04-16
Updating supported Python versions

rmeloamaro:
- As per the requirements listed in v0.5.0 Python WinRM plugin: https://github.com/diyan/pywinrm


fdevans:
I don't believe we are testing with these versions today so until that changes we wouldn't update the docs around it.

PR #1666, opened 2025-05-28
cleanup runner-install

gschueler:
remove "comments" which are leaking into the content


jsboak (approved):
get those comments outta here!

PR #1682, opened 2025-07-18
RUN-3515: Create CVE-2023-0833.md

smartinellibenedetti:
This CVE is a false positive: OkHttp was bundled in agent.jar because runner-agent's build.gradle uses the shadow plugin to create a fat JAR that includes all transitive dependencies, including the OkHttp library that comes via the Retrofit dependency from core. Neither runner-agent nor rundeckpro-runner plugin were using it, so the vulnerability wasn’t applicable in our case. Furthermore, the CVE disappeared when retrofit was updated.

### DIFF
--- a/docs/history/cves/CVE-2023-0833.md
+++ b/docs/history/cves/CVE-2023-0833.md
@@ -1,0 +1,16 @@
+---
+order: 51
+---
+
+# CVE-2023-0833
+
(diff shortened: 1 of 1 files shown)


smartinellibenedetti:
Closing PR as the CVE won't show up on scans after 5.14.0

PR #1704, opened 2025-09-09
Fixed key storage path - group-project-full.md

tknog-pd:
Changed to     

match:
      path: keys/project/prj-sandbox(/.*)?

Removed 
    name: .*


tknog-pd:
It seems that some error happened. Please review this request since the code for setting the key storage path is not working properly. Thank you.

fdevans (approved):

PR #1707, opened 2025-09-29
RUN-643: Multiline job options beta documentation

gschueler:
RUN-643

for https://github.com/rundeck/rundeck/pull/9822


smartinellibenedetti (dismissed):
Approved the docs, but as mentioned in the other PR would suggest to add a note that toggling the multiline job option might break the jobs that had an option of this type.

fdevans (approved):

fdevans:
@smartinellibenedetti I updated the docs with content based on your warning in this PR. https://github.com/rundeck/rundeck/pull/9822

PR #1716, opened 2025-10-27
RUN-2210: Update Key Value Data with new Match Substrings checkbox

edbaltra:
JIRA TICKET:https://pagerduty.atlassian.net/browse/RUN-2210

This pull request updates the documentation for log filter configuration components by introducing a new option and clarifying its behavior. The most significant change is the addition of the "Match Substrings" checkbox, which affects how patterns are matched in log output.

**Documentation update:**

* Added a description of the new `Match Substrings` checkbox to the list of configuration components, explaining that it allows the pattern to match substrings within a line of log output rather than requiring a full line match.

### DIFF
--- a/docs/manual/log-filters/key-value-data.md
+++ b/docs/manual/log-filters/key-value-data.md
@@ -15,12 +15,15 @@ There are only three configuration components:
   - Pattern
   - Name Data
   - Log Data (Checkbox)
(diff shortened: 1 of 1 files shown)


jayas006 (approved):
LGTM 🎉

PR #1725, opened 2025-11-12
Add Workflow for Latest Updates Generation

fdevans:
This workflow will be called from CircleCI when a new SaaS Release Cut is generated.  It will create a PR that would be merged once that cut is pushed to Production.


fdevans:
Have to merge to test workflow

PR #1761, opened 2026-01-27
[DRAFT-RUN-3402 for IR-455]: Add instance health checks documentation for Demo purpose

jayas006:
Documents existing health check endpoints and provides guidance for customers and Support to check instance health. Addresses gap identified in RUN-3402/IR-455.


fdevans:
I don't know that all of these are available in SaaS so we need to note where it's Self Hosted only.

PR #1810, opened 2026-05-28
Update Azure SSO to use least-privilege permissions

brmdias:
Replace Directory.Read.All with GroupMember.Read.All for group name mapping. Make Application permissions optional - only required when using group names instead of group IDs in ACL policies.

### DIFF
--- a/docs/administration/security/sso/azure-sso.md
+++ b/docs/administration/security/sso/azure-sso.md
@@ -6,7 +6,7 @@ Rundeck can be configured to use Azure Active Directory authentication by regist
 
 ### Azure - Create a new app registration
 
-The first thing to do is create a new application registration in Azure.
+The first thing to do is create a new application registration in Azure.\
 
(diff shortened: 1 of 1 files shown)


docs/administration/security/sso/azure-sso.md
```diff
@@ -6,7 +6,7 @@ Rundeck can be configured to use Azure Active Directory authentication by regist
 
 ### Azure - Create a new app registration
 
-The first thing to do is create a new application registration in Azure.
+The first thing to do is create a new application registration in Azure.\
```
fdevans: Why is the slash at the end of this sentence?

fdevans (changes requested):

docs/administration/security/sso/azure-sso.md
brmdias: It's a typo.
Let me remove it.

fdevans (approved):

